PR #2, opened 2023-03-13
fdb5 library needs to be shared always

wdeconinck:
Due to reliance on factory self registration this library should never be compiled as static.
This may happen otherwise when CMake configuration with `-DBUILD_SHARED_LIBS=OFF` occurs

### DIFF
--- a/src/fdb5/CMakeLists.txt
+++ b/src/fdb5/CMakeLists.txt
@@ -368,6 +368,8 @@ endif()
 ecbuild_add_library(
 
     TARGET  fdb5
+    
+    TYPE SHARED # Library needs to be shared for dynamic factory self registration
 
     SOURCES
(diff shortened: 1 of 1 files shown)


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/fdb?pullRequest=2) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/fdb?pullRequest=2) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/fdb?pullRequest=2) it.</sub>

PR #20, opened 2024-03-18
Proposal to create an FDB instance from config path via C-API (for pyfdb)

pgeier:
Related to pyfdb #


tbkr:
@geier1993 is this still relevant?

geier1993:
Think it was related to the pyfdb PR that we just closed...
I know that Tom has introduced something similar back then???

Is there something similar available now... without having to use the ENV variables to point to a fdb root???

I think back then we were writing the fdb compare tool....

tbkr:
I'll close this due to the implemented solution in pyfdb (which is able to construct a fdb from a config string). In case something like this is needed still, feel free to re-open.

PR #24, opened 2024-06-10
Feature/archive callback

ChrisspyB:
API for registering a callback function `foo(key, location)`, to be called upon archiving to the Store.


tests/fdb/api/test_archive_callback.cc
```diff
@@ -0,0 +1,73 @@
...
+	fdb.flush();
+
+    EXPECT(map.size() == 3);
+
+    // Print out the map
+    for (const auto& [key, uri] : map) {
+        std::cout << key << " -> " << uri << std::endl;
+    }
+
+    // Note that the keys are not the same, but they are equivalent
```
ChrisspyB, line 50: Comment no longer true

src/fdb5/database/ArchiveVisitor.cc
```diff
@@ -29,8 +30,7 @@ bool ArchiveVisitor::selectDatum(const Key &key, const Key &full) {
 
     ASSERT(current());
 
-    current()->archive(key, data_, size_);
-
+    current()->archive(key, data_, size_, field_, callback_); //field_ is the user key! (or rather, "internal key")
```
ChrisspyB, line 33: Get rid of this comment

src/fdb5/api/FDB.h
```diff
@@ -149,6 +154,8 @@ class FDB {
     bool reportStats_;
 
     FDBStats stats_;
+
+    ArchiveCallback callback_ = nullptr;
```
simondsmart, line 158: Why is this both here and in FDBBase. It feels like it should belong there more cleanly.

src/fdb5/api/FDBFactory.h
```diff
@@ -125,6 +128,8 @@ class FDBBase : private eckit::NonCopyable {
     ControlIdentifiers controlIdentifiers_;
 
     bool disabled_;
+
+    ArchiveCallback callback_;
```
simondsmart, line 132: Where is the default value for this being set? And should that be nullptr, or should it be a NOP function?

src/fdb5/api/FDB.h
```diff
@@ -69,13 +70,15 @@ class FDB {
 
     // -------------- Primary API functions ----------------------------
 
-    void archive(eckit::message::Message msg);
+    Key archive(eckit::message::Message msg);
```
simondsmart, line 73: This return type should be (optionally) available in the C API.

src/fdb5/api/FDB.h
```diff
@@ -116,6 +119,8 @@ class FDB {
 
     bool dirty() const;
 
+    void registerCallback(ArchiveCallback callback);
```
simondsmart, line 122: Do we want to push this functionality to the C API. If not, why not?

src/fdb5/api/LocalFDB.cc
```diff
@@ -46,14 +46,15 @@ using namespace eckit;
...
+
 void LocalFDB::archive(const Key& key, const void* data, size_t length) {
 
     if (!archiver_) {
         LOG_DEBUG_LIB(LibFdb5) << *this << ": Constructing new archiver" << std::endl;
         archiver_.reset(new Archiver(config_));
     }
 
-    archiver_->archive(key, data, length);
+    archiver_->archive(key, data, length, callback_);
```
simondsmart, line 57: It feels like the callback could be passed to the archiver at construction time. If necessary it could be updated? Or is there a reason to pass it in each time?

src/fdb5/api/FDB.cc
```diff
@@ -123,6 +123,7 @@ void FDB::archive(const Key& key, const void* data, size_t length) {
 
     timer.stop();
     stats_.addArchive(length, timer);
+    return keyInternal;
```
simondsmart, line 126: I don't think that this is correct. The registry has been attached here, but this key is still the pre-canonicalisation key.

You need to return from internal_->archive()

src/fdb5/database/ArchiveVisitor.h
```diff
@@ -43,6 +43,8 @@ class ArchiveVisitor : public BaseArchiveVisitor {
     const void *data_;
     size_t size_;
 
+    ArchiveCallback callback_;
```
simondsmart, line 46: Is there value in taking this by reference rather than by copy?

src/fdb5/database/BaseArchiveVisitor.h
```diff
@@ -46,12 +46,14 @@ class BaseArchiveVisitor : public WriteVisitor {
 
     fdb5::DB* current() const;
 
+protected: // members
+
+    const Key &field_;
```
simondsmart, line 51: Alignment. Surely "const Key& field_"

field_ has reference type, so & should be part of the type.

simondsmart: Where is this being used?

src/fdb5/database/DB.cc
```diff
@@ -103,13 +104,19 @@ eckit::DataHandle *DB::retrieve(const Key& key) {
...
+void DB::archive(const Key& key, const void* data, eckit::Length length, const Key& internalKey, ArchiveCallback callback) {
 
     CatalogueWriter* cat = dynamic_cast<CatalogueWriter*>(catalogue_.get());
     ASSERT(cat);
 
     const Index& idx = cat->currentIndex();
-    cat->archive(key, store().archive(idx.key(), data, length));
+    std::unique_ptr<FieldLocation> location(store().archive(idx.key(), data, length));
+
+    if (callback) {
```
simondsmart, line 115: This makes me wonder if we do better to have a NOP function, rather than a control-flow if() in the pathway.

tests/fdb/api/test_archive_callback.cc
```diff
@@ -0,0 +1,73 @@
...
+        map[internalKey] = location.fullUri();
+    });
+
+	key.set("step","1");
+    internalKeys.push_back(fdb.archive(key, data, length));
+
+	key.set("step","2");
+    internalKeys.push_back(fdb.archive(key, data, length));
+
+	key.set("step","3");
```
simondsmart, line 38: Please make sure we have a test that checks changing metadata at level 1/2 of the schema as well. That has different internal impacts.

tests/fdb/api/test_archive_callback.cc
```diff
@@ -0,0 +1,73 @@
...
+	key.set("param","130");
+
+    std::map<fdb5::Key, eckit::URI> map;
+    std::vector<Key> internalKeys;
+
+    fdb.registerCallback([&map] (const fdb5::Key& internalKey, const fdb5::FieldLocation& location) {
+        map[internalKey] = location.fullUri();
+    });
+
+	key.set("step","1");
```
simondsmart, line 32: Why is the indentation different here?

tests/fdb/api/test_archive_callback.cc
```diff
@@ -0,0 +1,73 @@
...
+
+	key.set("step","3");
+    internalKeys.push_back(fdb.archive(key, data, length));
+    
+	fdb.flush();
+
+    EXPECT(map.size() == 3);
+
+    // Print out the map
+    for (const auto& [key, uri] : map) {
```
simondsmart, line 46: We don't want spurious output in (passing) tests. That's for debugging (you can leave commented if useful).

simondsmart (changes requested):
Some minor tweaks. Please see comments.

tests/fdb/api/test_archive_callback.cc
ChrisspyB: I reckoned it was okay since `ctest` will not actually print stdout unless you specify `--verbose`, but I will comment out

src/fdb5/api/LocalFDB.cc
ChrisspyB: I think you are right

src/fdb5/api/FDB.h
ChrisspyB: Likely a mistake

src/fdb5/api/FDB.cc
ChrisspyB: I agree this is not the canonicalised key. I'll look into this again, but I think it might not be possible to return the fully canonicalised key due to it only being created in the recursive traversal of the schema.

src/fdb5/database/BaseArchiveVisitor.h
ChrisspyB: The alignment is simply what was there before. This was previous a private member, that Ive promoted to protected so the archive visitor can give it to DB, which passes it to the callback.

ChrisspyB: Note, field_ is a non-canonical key.

Honestly, the treatment of keys is very unclear throughout. You never know when something is canonical, or just a keypart, etc. etc.

src/fdb5/api/FDBFactory.h
ChrisspyB: Added a NOP function

src/fdb5/api/LocalFDB.cc
ChrisspyB: Moved to constructor

src/fdb5/database/ArchiveVisitor.h
ChrisspyB: Probably ought to be using reference essentially everywhere, updated to do that.

ChrisspyB:
We decided to not return a key from the archive method, and also decided not to bother propgating this change to the C API just yet

PR #28, opened 2024-07-15
Feature/gribjump plugin

ChrisspyB:


ChrisspyB:
**NB** In case I forget: requires new `setup` method in eckit::system::plugin to be merged into eckit.

All tests will fail to build until then.

src/fdb5/api/FDB.h
```diff
@@ -120,6 +120,7 @@ class FDB {
     bool dirty() const;
 
     void registerCallback(ArchiveCallback callback);
+    void registerCallback(FlushCallback callback);
```
simondsmart, line 123: registerFlushCallback()

Consider the case of a lambda or a nop function. Rename.

src/fdb5/api/FDB.h
```diff
@@ -155,6 +158,7 @@ class FDB {
 
     FDBStats stats_;
 
+    FlushCallback flushCallback_ = CALLBACK_FLUSH_NOOP;
```
simondsmart, line 161: In constructor:

flushCallback_ = LibFDB::instance().defaultFlushCallback();

src/fdb5/api/FDB.cc
```diff
@@ -325,6 +331,19 @@ void FDB::registerCallback(ArchiveCallback callback) {
     internal_->registerCallback(callback);
 }
 
+void FDB::registerCallback(FlushCallback callback) {
+    flushCallback_ = callback;
+}
+
+void FDB::initPlugins(const Config& config){
+    bool enableGribjump = eckit::Resource<bool>("fdbEnableGribjump;$FDB_ENABLE_GRIBJUMP", false);
```
simondsmart, line 339: I do not want to see the words gribJump in the FDB (or really even the word GRIB)

src/fdb5/api/FDB.h
```diff
@@ -146,6 +147,8 @@ class FDB {
 
     bool sorted(const metkit::mars::MarsRequest &request);
 
+    void initPlugins(const Config& config);
```
simondsmart, line 150: Nuke this function.

src/fdb5/api/helpers/ArchiveCallback.h
```diff
@@ -20,8 +20,10 @@
 
 namespace fdb5 {
 
-using ArchiveCallback = std::function<void(const Key&, const FieldLocation&)>;
+using ArchiveCallback = std::function<void(const Key& key, const void* data, size_t length, const FieldLocation&)>;
+using FlushCallback = std::function<void()>;
 
-static const ArchiveCallback CALLBACK_NOOP = [](const Key&, const FieldLocation&) {};
+static const ArchiveCallback CALLBACK_NOOP = [](const Key& key, const void* data, size_t length, const FieldLocation&) {};
+static const FlushCallback CALLBACK_FLUSH_NOOP = []() {};
```
simondsmart, line 27: This file is called ArchiveCallback.h - clearly wrong if FlushCallback is there.

src/fdb5/api/helpers/ArchiveCallback.h
```diff
@@ -20,8 +20,10 @@
 
 namespace fdb5 {
 
-using ArchiveCallback = std::function<void(const Key&, const FieldLocation&)>;
+using ArchiveCallback = std::function<void(const Key& key, const void* data, size_t length, const FieldLocation&)>;
```
simondsmart, line 23: Pass the data and a promise<FieldLocation>
Note in comment that the callback must return _before_ the FieldLocation will be satisfied (so promise.get() will need to be called in a thread/async/something)

ChrisspyB: Passing a future rather than the promise (promise allows you to set the value, future allows you to get the value. Callback should only get the value).

> callback must return before the FieldLocation will be satisfied

Personally, I think it would be better if the callback did not *have* to implemented its own thread (though we do for gribjump): the important thing is that promse.set_value(location) is called before calling the callback (when using a single-threaded fdb backend), or the thread which sets the value has been started (when using remote fdb) before calling the callback.

simondsmart:
Looks great. Let's do it.

PR #30, opened 2024-08-01
list feature: key levels

mcakircali:
adds level var to fdb::list

new fdb-list option: --depth

cherry-picked (from Simon's other branch) and fixed some issues

### DIFF
--- a/src/fdb5/api/FDB.cc
+++ b/src/fdb5/api/FDB.cc
@@ -169,7 +169,7 @@ eckit::DataHandle* FDB::read(const std::vector<eckit::URI>& uris, bool sorted) {
     }
     return result.dataHandle();
 }
-    
+
 
 eckit::DataHandle* FDB::read(ListIterator& it, bool sorted) {
     eckit::Timer timer;
@@ -231,8 +231,8 @@ ListIterator FDB::inspect(const metkit::mars::MarsRequest& request) {
     return internal_->inspect(request);
 }
 
-ListIterator FDB::list(const FDBToolRequest& request, bool deduplicate) {
-    return ListIterator(internal_->list(request), deduplicate);
+ListIterator FDB::list(const FDBToolRequest& request, const bool deduplicate, const int level) {
+    return {internal_->list(request, level), deduplicate};
 }
 
 DumpIterator FDB::dump(const FDBToolRequest& request, bool simple) {
--- a/src/fdb5/api/local/ListVisitor.h
+++ b/src/fdb5/api/local/ListVisitor.h
@@ -25,9 +25,7 @@
 #include "fdb5/api/helpers/ListIterator.h"
 #include "eckit/os/BackTrace.h"
 
-namespace fdb5 {
-namespace api {
-namespace local {
+namespace fdb5::api::local {
 
 /// @note Helper classes for LocalFDB
 
@@ -58,8 +56,9 @@ public:
         }
 
         if (level_ == 1) {
-            eckit::Log::info() << eckit::BackTrace::dump() << std::endl;
-            queue_.emplace(ListElement({currentCatalogue_->key(), Key(), Key()}, FieldLocation::nullLocation(), 0));
+            /// @todo fix this so that we don't need to create an empty key
+            const auto emptyKey = Key {currentCatalogue_->schema().registry()};
+            queue_.emplace(ListElement({currentCatalogue_->key(), emptyKey, emptyKey}, FieldLocation::nullLocation(), 0));
             ret = false;
         }
 
@@ -82,7 +81,9 @@ public:
 
         if (index.partialMatch(request_)) {
             if (level_ == 2) {
-                const auto keyParts = std::vector<Key> {currentCatalogue_->key(), currentIndex_->key(), Key()};
+                /// @todo fix this so that we don't need to create an empty key
+                const auto emptyKey = Key {currentCatalogue_->schema().registry()};
+                const auto keyParts = std::vector<Key> {currentCatalogue_->key(), currentIndex_->key(), emptyKey};
                 queue_.emplace(ListElement(keyParts, FieldLocation::nullLocation(), 0));
                 return false;
             }
@@ -115,8 +116,6 @@ private: // members
 
 //----------------------------------------------------------------------------------------------------------------------
 
-} // namespace local
-} // namespace api
-} // namespace fdb5
+}  // namespace fdb5::api::local
 
 #endif
--- a/src/fdb5/api/local/ListVisitor.h
+++ b/src/fdb5/api/local/ListVisitor.h
@@ -23,6 +23,7 @@
 #include "fdb5/database/Index.h"
 #include "fdb5/api/local/QueryVisitor.h"
 #include "fdb5/api/helpers/ListIterator.h"
+#include "eckit/os/BackTrace.h"
 
 namespace fdb5 {
 namespace api {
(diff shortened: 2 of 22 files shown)


src/fdb5/api/local/ListVisitor.h
```diff
@@ -55,6 +55,13 @@ struct ListVisitor : public QueryVisitor<ListElement> {
             indexRequest_.unsetValues(kv.first);
         }
 
+        if (level_ == 1) {
+            /// @todo fix this so that we don't need to create an empty key
+            const auto emptyKey = Key {currentCatalogue_->schema().registry()};
```
simondsmart, line 60: Is this necessary with the schema().registry() - given that there are no keys contained inside? Does anything break if we just default construct the empty keys().

Further, can we not just emplace a single part key here, rather than all 3 parts?

src/fdb5/database/Manager.cc
```diff
@@ -298,7 +298,7 @@ std::vector<eckit::URI> Manager::visitableLocations(const metkit::mars::MarsRequ
 
     std::set<std::string> engines = Manager::engines(rq, all);
 
-    LOG_DEBUG_LIB(LibFdb5) << "Matching engines for request " << rq << " -> " << engines << std::endl;
+    LOG_DEBUG_LIB(LibFdb5) << "Matching engines for request " << rq << (all ? " ALL" : "") << " -> " << engines << std::endl;
```
simondsmart, line 301: Not quite sure what the value of this change is?

src/fdb5/remote/Handler.cc
```diff
@@ -88,7 +88,10 @@ struct BaseHelper {
 };
 
 struct ListHelper : public BaseHelper<ListElement> {
+    // void extraDecode(eckit::Stream& s) { s >> level_; }
```
simondsmart, line 91: This is incomplete. As discussed, the remote side needs to work...

src/fdb5/tools/fdb-list.cc
```diff
@@ -29,54 +29,47 @@
...
         options_.push_back(new SimpleOption<bool>("timestamp", "Also print the timestamp when the field was indexed"));
         options_.push_back(new SimpleOption<bool>("length", "Also print the field size"));
         options_.push_back(new SimpleOption<bool>("full", "Include all entries (including masked duplicates)"));
-        options_.push_back(new SimpleOption<bool>("porcelain", "Streamlined and stable output for input into other tools"));
+        options_.push_back(new SimpleOption<bool>("porcelain",
+                                                  "Streamlined and stable output. "
+                                                  "Useful as input for other tools or scripts"));
         options_.push_back(new SimpleOption<bool>("json", "Output available fields in JSON form"));
         options_.push_back(new SimpleOption<bool>("compact", "Aggregate available fields in MARS requests"));
+        options_.push_back(new SimpleOption<long>("depth", "Output entries up to 'depth' levels deep"));
```
simondsmart, line 48: Note " (default 3)" in the usage statement.

src/fdb5/tools/fdb-list.cc
```diff
@@ -29,54 +29,47 @@
...
 
+//----------------------------------------------------------------------------------------------------------------------
 
 std::string keySignature(const fdb5::Key& key) {
     std::string signature;
     std::string separator;
-    for (auto k : key.keys()) {
-        signature += separator+k;
-        separator=":";
+    for (auto&& k : key.keys()) {
```
simondsmart, line 70: Should be "const auto&". Doesn't need && - not intending either to do move(), perfect forwarding or lifetime extension.

src/fdb5/tools/fdb-list.cc
```diff
@@ -93,24 +86,20 @@ void FDBList::init(const CmdArgs& args) {
     porcelain_ = args.getBool("porcelain", false);
     json_ = args.getBool("json", false);
     compact_ = args.getBool("compact", false);
+    depth_     = args.getInt("depth", 3);
```
simondsmart, line 89: Given that we set the default in the class/constructor, we should default getInt("depth", depth_). That probably applies for the other arguments as well...

src/fdb5/tools/fdb-list.cc
```diff
@@ -93,24 +86,20 @@ void FDBList::init(const CmdArgs& args) {
...
 
-    if (json_) {
-        porcelain_ = true;
-        if (location_ || timestamp_ || length_) {
-            throw UserError("--json and --location/--timestamp/--length not compatible", Here());
-        }
+    if (json_) { porcelain_ = true; }
+
+    if (porcelain_) {
```
simondsmart, line 93: This is quite elegant, in that it removes the need for if (porcelain_) later on.

But I would rather that we still throw a UserError if the user asks for something else, and explain what the conflict is, rather than just give them an output that they likely don't expect.

simondsmart (changes requested):
I would like you to review this in two main ways:
* Please don't refactor tests in the regressions folder. I would like to see what fails (and what the minimum possible changes are) to make FDB-240 work. That _should not break_ with any of the changes in this PR
* I don't like the output that tails empty {} in the list output. It isn't obvious that the ListElement should always contain 3 components when level < 3.
* The RemoteFDB handler needs to be made to work
* Did you look at what would be involved in removing the NullFieldLocation?

src/fdb5/database/Manager.cc
mcakircali: I can't tell if useful or not, it was cherry-picked from your branch.

src/fdb5/tools/fdb-list.cc
mcakircali: unrelated perf fix.. here, "auto&&" becomes "const auto&"

src/fdb5/api/local/ListVisitor.h
mcakircali: we need this hacky empty keys since there's a Key hash functor using valuesToString .. which kicks in when Key used in unordered containers.. see the note, I will attempt to fix in a later PR.

src/fdb5/api/FDB.cc
```diff
@@ -231,8 +231,8 @@ ListIterator FDB::inspect(const metkit::mars::MarsRequest& request) {
     return internal_->inspect(request);
 }
 
-ListIterator FDB::list(const FDBToolRequest& request, bool deduplicate) {
-    return ListIterator(internal_->list(request), deduplicate);
+ListIterator FDB::list(const FDBToolRequest& request, const bool deduplicate, const int level) {
```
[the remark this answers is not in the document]
mcakircali: no, it's correctly set as const. it's passed by value to ListIterator, where it'll live as a non-const.

mcakircali:
> I would like you to review this in two main ways:
> 
> * Please don't refactor tests in the regressions folder. I would like to see what fails (and what the minimum possible changes are) to make FDB-240 work. That _should not break_ with any of the changes in this PR

that's fine, I can understand it. I'm not anxious since I reviewed every line but will remove that commit.

> * I don't like the output that tails empty {} in the list output. It isn't obvious that the ListElement should always contain 3 components when level < 3.

this is already WIP. there will be a separate PR since it's a big change.

> * The RemoteFDB handler needs to be made to work

Good point, I will fix that.

> * Did you look at what would be involved in removing the NullFieldLocation?

Yes, as mentioned, in a separate PR.

src/fdb5/tools/fdb-list.cc
mcakircali: good point

mcakircali:
closing, see more involved PR https://github.com/ecmwf/fdb/pull/31

PR #38, opened 2024-09-18
fix(Store): archive returns const FieldLocation

mcakircali:


danovaro:
this looks uncontroversial

PR #48, opened 2024-10-28
rule schema key

mcakircali:


src/fdb5/tools/fdb-list.cc
```diff
@@ -8,106 +8,110 @@
...
         options_.push_back(new SimpleOption<bool>("location", "Also print the location of each field"));
         options_.push_back(new SimpleOption<bool>("timestamp", "Also print the timestamp when the field was indexed"));
         options_.push_back(new SimpleOption<bool>("length", "Also print the field size"));
         options_.push_back(new SimpleOption<bool>("full", "Include all entries (including masked duplicates)"));
-        options_.push_back(new SimpleOption<bool>("porcelain", "Streamlined and stable output for input into other tools"));
+        options_.push_back(
+            new SimpleOption<bool>("porcelain",
+                                   "Streamlined and stable output. Useful as input for other tools or scripts."
+                                   "Incompatible with options: location, timestamp, and length"));
         options_.push_back(new SimpleOption<bool>("json", "Output available fields in JSON form"));
```
Ozaq, line 54: Since you already updated the doc: This also needs to mention the interaction of `--json` with `--porcelain` and the resulting incompatibilities

PR #51, opened 2024-11-19
Feature/c header

danovaro:


src/dummy_daos/daos.h
```diff
@@ -16,12 +16,12 @@
 
 #pragma once
 
-#include <stdlib.h>
-#include <stdint.h>
+#include <cstdlib>
```
Ozaq, line 19: The file is written as a C header, i.e. it features a `ifdef __cplusplus`, changing the c headers to c++ headers will break the c compatibility. 

This file probably does not need to expose a C-API, but in that case the Ifdefs should also be removed as they give the impression of C-compatiblity

src/dummy_daos/daos.h
```diff
@@ -16,12 +16,12 @@
...
+#include <cstdint>
 #include <unistd.h>
 #include <uuid/uuid.h>
-#include <stdbool.h>
-#include <string.h>
+#include <cstdbool>
```
Ozaq, line 23: This one is deprecated with C++17 and should be removed as c++ has true/false as keywords

src/fdb5/api/RandomFDB.cc
```diff
@@ -19,7 +19,7 @@
 #include "fdb5/api/FDBFactory.h"
 #include "fdb5/LibFdb5.h"
 
-#include <stdlib.h>
+#include <cstdlib>
```
Ozaq, line 22: Can be removed it is unused

src/fdb5/database/Archiver.h
```diff
@@ -16,7 +16,7 @@
 #ifndef fdb5_Archiver_H
 #define fdb5_Archiver_H
 
-#include <time.h>
+#include <ctime>
```
Ozaq, line 19: It is unused in the header and only used in the corresponding .cc, should be moved into that.

src/fdb5/remote/FdbServer.h
```diff
@@ -16,7 +16,7 @@
 #define fdb5_remote_FdbServer_H
 
 #include <unistd.h>
-#include <stdlib.h>
+#include <cstdlib>
```
Ozaq, line 19: Not used here directly, needs to be moved in to the .cc file

src/fdb5/toc/TocMoveVisitor.cc
```diff
@@ -22,8 +22,7 @@
 #include "fdb5/toc/TocMoveVisitor.h"
 #include "fdb5/toc/RootManager.h"
 
-#include <dirent.h>
```
Ozaq, line 25: dirent is used. This sill works because this header is transitively included by other headers but IMO it should be kept here.

src/fdb5/toc/TocMoveVisitor.cc
```diff
@@ -22,8 +22,7 @@
...
-#include <errno.h>
+#include <cerrno>
```
Ozaq, line 25: This can be removed, it is unused in this file

tests/fdb/type/test_toKey.cc
```diff
@@ -9,7 +9,7 @@
  */
 
 #include <cstdlib>
-#include <string.h>
+#include <cstring>
```
Ozaq, line 12: Unused in this file

PR #63, opened 2025-01-06
Add callback interface class for serverside callback registration

ChrisspyB:
Simple abstraction of the callback registration to allow registration of callbacks with the StoreHandler, rather than just an FDB object. Required for gribjump plugin on the databridge store servers, which at no point create an FDB object.

### DIFF
--- a/src/fdb5/api/helpers/Callback.h
+++ b/src/fdb5/api/helpers/Callback.h
@@ -21,13 +21,30 @@
 namespace fdb5 {
 
 class FDB;
+class CallbackInterface;
 
 using ArchiveCallback = std::function<void(const Key& key, const void* data, size_t length, std::future<std::shared_ptr<const FieldLocation>>)>;
 using FlushCallback = std::function<void()>;
-using ConstructorCallback = std::function<void(FDB&)>;
+using ConstructorCallback = std::function<void(CallbackInterface&)>;
 
-static const ArchiveCallback CALLBACK_NOOP = [](const Key& key, const void* data, size_t length, std::future<std::shared_ptr<const FieldLocation>>) {};
+static const ArchiveCallback CALLBACK_ARCHIVE_NOOP = [](const Key& key, const void* data, size_t length, std::future<std::shared_ptr<const FieldLocation>>) {};
 static const FlushCallback CALLBACK_FLUSH_NOOP = []() {};
-static const ConstructorCallback CALLBACK_CONSTRUCTOR_NOOP = [](FDB&) {};
+static const ConstructorCallback CALLBACK_CONSTRUCTOR_NOOP = [](CallbackInterface&) {};
+
+// -------------------------------------------------------------------------------------------------
+
+// This class provides a common interface for registering callbacks with an FDB object or a Store/Catalogue Handler.
+class CallbackInterface {
+public:
+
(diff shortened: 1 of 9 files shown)


ChrisspyB:
Related: https://github.com/ecmwf/gribjump/pull/37

ChrisspyB:
Not ready for merge yet, some testing to do. But I am optimisitic.

ChrisspyB:
Seems to work, ready to review

src/fdb5/api/helpers/Callback.h
```diff
@@ -21,13 +21,30 @@
...
 using ArchiveCallback = std::function<void(const Key& key, const void* data, size_t length, std::future<std::shared_ptr<const FieldLocation>>)>;
 using FlushCallback = std::function<void()>;
-using ConstructorCallback = std::function<void(FDB&)>;
+using ConstructorCallback = std::function<void(CallbackInterface&)>;
 
-static const ArchiveCallback CALLBACK_NOOP = [](const Key& key, const void* data, size_t length, std::future<std::shared_ptr<const FieldLocation>>) {};
+static const ArchiveCallback CALLBACK_ARCHIVE_NOOP = [](const Key& key, const void* data, size_t length, std::future<std::shared_ptr<const FieldLocation>>) {};
 static const FlushCallback CALLBACK_FLUSH_NOOP = []() {};
-static const ConstructorCallback CALLBACK_CONSTRUCTOR_NOOP = [](FDB&) {};
+static const ConstructorCallback CALLBACK_CONSTRUCTOR_NOOP = [](CallbackInterface&) {};
```
Ozaq, line 32: You can omit the types in the parameters or the `NOOP` CBs, e.g.

```c++
static const ArchiveCallback CALLBACK_ARCHIVE_NOOP = [](auto&&, auto&&, auto&&, auto&&) {};
 static const FlushCallback CALLBACK_FLUSH_NOOP = [](){};
 static const ConstructorCallback CALLBACK_CONSTRUCTOR_NOOP = [](auto&&) {};
```

Note you need to use `auto&&` to prevent introducing a copy for types passed as references when calling the `NOOP`

src/fdb5/api/helpers/Callback.h
```diff
@@ -21,13 +21,30 @@
...
+
+// -------------------------------------------------------------------------------------------------
+
+// This class provides a common interface for registering callbacks with an FDB object or a Store/Catalogue Handler.
+class CallbackInterface {
```
Ozaq, line 37: This is not an "Interface" as it has state and a default implementation and the name is a bit too generic. How about something along the lines of "PersistanceNotification" because the CBs are called on data persistence operations?

Also I think It would be better to not provide half the implementation in this class, i.e. storage / initialization of the callbacks and actual usage in the derived classes. Code duplication of the assignment is clearer in my opinion, which would leave you with a "real" interface and a tiny bit of similar code in FDBBase & StoreHandler

ChrisspyB: > not an "Interface"
Agree. I will rename to registry, which is not a great name, but at least a bit more accurate.

> "PersistanceNotification" because the CBs are called on data persistence operations?
I think we would add to this class down the line if we add callbacks elsewhere, so I think a slightly more generic name is better.

PR #67, opened 2025-01-14
remoteFDB - constness improvements

mcakircali:
closed see #68

### DIFF
--- a/src/fdb5/remote/Connection.cc
+++ b/src/fdb5/remote/Connection.cc
@@ -4,6 +4,9 @@
 #include "fdb5/LibFdb5.h"
 #include "fdb5/remote/Connection.h"
 #include "fdb5/remote/Messages.h"
+#include <cstdint>
+#include <mutex>
+#include <string_view>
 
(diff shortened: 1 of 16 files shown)


src/fdb5/remote/Connection.cc
```diff
@@ -19,20 +19,20 @@ void Connection::teardown() {
...
     } catch(...) {
-        // if connection is already down, no need to escalate 
+        // if connection is already down, no need to escalate
     }
 }
 
 //----------------------------------------------------------------------------------------------------------------------
 
-void Connection::writeUnsafe(bool control, const void* data, size_t length) {
+void Connection::writeUnsafe(const bool control, const void* data, const size_t length) const {
```
Ozaq, line 35: If you make all parameters const you should also make the pointer const, i.e. `const void * const`

mcakircali: there is no **should**, and feel free to contribute it as a commit!

mcakircali:
closing see #68

PR #68, opened 2025-01-16
Adds functionality to the client-side exists() methods

mcakircali:
given dbkey by client, client-server comm that checks catalogue/store exists on server
(includes the needed const-ness improvements)


danovaro (approved):
looks good to me, thanks @mcakircali

PR #78, opened 2025-02-11
Switch to reusable action

tmi:
Switches to reusable action, to have releases for all python versions and platforms

There are two caveats here:
- this is realistically useful only after https://github.com/ecmwf/metkit/pull/55/files gets merged
- the name of the resulting wheel is `fdblib`. That doesn't play well with `findlibs.find("fdb5")` (which is currently used in `pyfdb`), and instead `findlibs.find("fdb5", "fdblib")` is required. _Or_, we can instead name the wheel `fdb5lib` -- whatever you guys find suitable


Ozaq:
Id prefer if the name would match the wrapped .so, i.e. `libfdb5` and ofc this males most sense if this is applied to every wrapper python module. Just my 2 cents :)

Ozaq (approved):

PR #80, opened 2025-02-12
[Doc] Documentation for FDB.h

Ozaq:
This PR adds doc to FDB.h.

There are several TODOs added to mark a few yet undocumented methods.

Last time we opened this PR we tried to complete all of the doc in one go and could not finish it up in a reasonable time. This time I want to get this merged and I will address the open TODOs over the next couple of weeks as low priority work.


src/fdb5/api/FDB.h
```diff
@@ -72,73 +76,200 @@ class FDB {
     FDB(FDB&&)            = default;
     FDB& operator=(FDB&&) = default;
 
-    // -------------- Primary API functions ----------------------------
+    // -------------- Primary API functions ----------------------------------------------------------------------------
 
+    /// Archive binary data to a FDB.
+    /// Any callback set with registerArchiveCallback will be invoked.
+    /// @param handle eckit::message::Message to data to archive
```
simondsmart, line 83: Not just binary data. An eckit::Message.

This is data which has been pre-chunked into the objects to be stored. And that we know is self-describing (hence not needing a key to be supplied)

src/fdb5/api/FDB.h
```diff
@@ -72,73 +76,200 @@ class FDB {
...
     void archive(eckit::message::Message msg);
+
+    /// Archive binary data to a FDB.
+    /// Reads messages from the eckit::DataHandle and calls archive() the the
+    /// corresponding messages.
+    /// Any callback set with registerArchiveCallback will be invoked on each message.
+    /// @param handle eckit::DataHandle reference data to archive
```
simondsmart, line 90: Archives a stream of potentially multiple messages.

src/fdb5/api/FDB.h
```diff
@@ -72,73 +76,200 @@ class FDB {
...
 
-    // disclaimer: this is a low-level API. The provided key and the corresponding data are not checked for consistency
-    // Optional callback function is called upon receiving field location from the store.
+    /// Archive binary data to a FDB.
+    /// Any callback set with registerArchiveCallback will be invoked.
+    /// @note No constistency checks are applied. The caller needs to ensure the provided key matches metadata present
+    /// in data.
+    /// @param key Key used for indexing and archiving the data
+    /// @param data Pointer to the binary data to archive
+    /// @param length Size of the data to archive with the given @p key
```
simondsmart, line 114: Archives a single object.

Not sure I like the term 'binary data here' - we don't care what is in the blob. It is a blob. Call it an object, or similar.

src/fdb5/api/FDB.h
```diff
@@ -72,73 +76,200 @@ class FDB {
...
     void reindex(const Key& key, const FieldLocation& location);
 
-    /// Flushes all buffers and closes all data handles into a consistent DB state
+    /// Flush all buffers and closes all data handles into a consistent DB state
     /// @note always safe to call
     void flush();
 
+    /// Read binary data from an URI.
+    /// @param uri eckit uri to the data source
+    /// @return DataHandle for reading the requested data from
```
simondsmart, line 129: Add note to review. This is a bit odd. The purpose of a URI is that it directly describes the data locations - and as such shouldn't need the FDB object to do the conversion into DataHandle?

src/fdb5/api/FDB.h
```diff
@@ -72,73 +76,200 @@ class FDB {
...
+    /// the order it was requested.
+    /// @return DataHandle for reading the requested data from
+    eckit::DataHandle* read(ListIterator& it, bool inStorageOrder = false);
+
+    /// Retrieve data which is specified by a MARS request.
+    /// @param request MarsRequest which describes the data which should be retrieved
+    /// @return DataHandle for reading the requested data from
     eckit::DataHandle* retrieve(const metkit::mars::MarsRequest& request);
 
+    // TODO(kkratz): Provide doc!
```
simondsmart, line 151: Similar to list, but walks the schema rather than walking the contents.

Optimised for small retrieve against a large dataset. Used as part of the internal of the retrieve() call.

src/fdb5/api/FDB.h
```diff
@@ -72,73 +76,200 @@ class FDB {
...
 
+    /// Move content of one FDB database.
+    /// This locks the source database, make it possible to create a second
+    /// database in another root, duplicates all data. Source data are not automatically removed.
+    /// @param request a fdb tool request for the data which should be move
+    /// @param dest destination uri to which the data should be moved
+    /// @return MoveIterator for iterating over the set of found items
     MoveIterator move(const FDBToolRequest& request, const eckit::URI& dest);
 
+    /// Remove duplicate data from the database.
```
simondsmart, line 196: General comment, that these need some 'vertical whitespace' (I guess blank /// lines). This is a bit of a 'block' of text.

simondsmart: It's good to have a clear one-line summary of the routine, and then detail laid out below if needed.

simondsmart (changes requested):

src/fdb5/api/FDB.h
Ozaq: done

src/fdb5/api/FDB.h
Ozaq: done

src/fdb5/api/FDB.h
Ozaq: done

PR #81, opened 2025-02-12
Remove internalStats from FDB.h and FDBBase

Ozaq:
This PR removes some unused functionality from FDB.h / internal classes.

Internal stats was always returning a empty default initialised FDBStats object. In case of using a RemoteFDB it would even throw a not implemnted exception.


tbkr:
LGTM

danovaro (approved):
looks good, thanks @Ozaq !

PR #82, opened 2025-02-12
Use downstream ci clang format

Ozaq:
downstream-ci is updated to use clang-format-19, see https://github.com/ecmwf-actions/downstream-ci/pull/83

Also clang-19 is now available on ATOS.

We can now use the default check-format action.


tbkr:
LGTM

PR #85, opened 2025-02-14
improved MarsRequest canonicalisation + memoization

danovaro:
canonical MarsRequest with value deduplication
improved canonical MarsRequest memoization
little cleanup


src/fdb5/types/TypesRegistry.h
```diff
@@ -83,4 +87,9 @@ class TypesRegistry : public eckit::Streamable {
 
 }  // namespace fdb5
 
+template <>
+struct std::hash<const fdb5::TypesRegistry*> {
+    std::size_t operator()(const fdb5::TypesRegistry* registry) const { return registry->hash(); }
+};
+
```
ChrisspyB, line 94: IIRC from our discussion, now that you have EqualsTypesRegistry and HashTypesRegistry in the QueryVisitor, this code coes not run? If so we should remove it.

src/fdb5/types/TypesRegistry.cc
```diff
@@ -98,11 +112,16 @@ metkit::mars::MarsRequest TypesRegistry::canonicalise(const metkit::mars::MarsRe
 
     for (const auto& param : request.parameters()) {
         const std::vector<std::string>& srcVals = param.values();
+        std::set<std::string> uniqueVals;
         std::vector<std::string> vals;
-        vals.reserve(srcVals.size());
         const Type& type = lookupType(param.name());
         for (const auto& v : srcVals) {
-            vals.push_back(type.toKey(v));
+            auto newVal = type.toKey(v);
```
ChrisspyB, line 119: Personally would use std::string newVal instead of auto

src/fdb5/toc/TocHandler.cc
```diff
@@ -1251,7 +1251,7 @@ const eckit::LocalPathName& TocHandler::directory() const {
     return directory_;
 }
 
-std::vector<Index> TocHandler::loadIndexes(const Catalogue& catalogue, bool sorted, std::set<std::string>* subTocs,
+std::vector<Index> TocHandler::loadIndexes(bool sorted, std::set<std::string>* subTocs,
```
ChrisspyB, line 1254: If I understand correctly, catalogue was passed around a lot in the TocHandler but never used?

ChrisspyB (approved):
Minor comments, lgtm.

src/fdb5/toc/TocHandler.cc
danovaro: exactly! Initially we added a reference to the catalogue (probably to reach the schema/rule).
With the improvements in Key, we no longer need the rule to build a canonicalised Key.
Now we are slowly removing the unnecessary dependencies

src/fdb5/types/TypesRegistry.h
danovaro: Instead of removing this, I've added an implementation of
struct std::equal_to<const fdb5::TypesRegistry*>
and removed both EqualsTypesRegistry and HashTypesRegistry. Before it was a hybrid between the two approaches

mcakircali (approved):
very nice. my comments are just alternatives and otherwise should be ignored

PR #86, opened 2025-02-17
Enable local build of doc incl. c/c++ api

Ozaq:
To enable local documentation build set cmake option
    FDB5_ENABLE_DOCUMENTATION=ON
This will require Doxygen 1.13 and sphinx and its requirements. Sphinx and deps can installed via pip from docs/requirements.txt

After succefull cmake configuration a new target fdb-doc will be added. This new target it part of 'all'.


docs/conf.py
```diff
@@ -1,123 +1,43 @@
...
-# Generate Doxygen documentation in the XML format.
-#assert subprocess.check_call("doxygen Doxyfile.in", stdout=subprocess.DEVNULL, shell=True) == 0
-
-
 # -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
-project = "fdb"
+project = "Fields DataBase - FDB"
+copyright = "2025, ECMWF"
```
tbkr, line 10: The moving copyright year was removed here. Shall we keep it?

tbkr (commented):
In general:
Let's discuss about the integration into CICD.
Otherwise nice and clean PR.

docs/conf.py
Ozaq: What's the correct year to start with?

PR #88, opened 2025-02-19
Fix misleading 'serious error' message

Ozaq:
During destruction of an AsyncApiIterator an SeriousBug exception is created if the writer task has not yet pushd all output into the shared eckit::Queue. Sending a SeriousBug exception to the writing thread, via eckit::Queue::interrupt, safely stops writing by unwinding the stack until the SeriousBug exception was silently caught in the wrapping lambda.

The issue has been that on construction of a SeriousBug exception error message and stack trace are logged. This implies to the user that a serious defect was encountered when in fact everything is fine.

This change is a bit of a hack as it still uses throwing an exception from the queue on write but now will use an exception that does not log anything.

Fixes: FDB-405


ChrisspyB (commented):
Does seem a bit hacky... I think I'll defer to Simon for this one.

Ozaq:
> Does seem a bit hacky... I think I'll defer to Simon for this one.

Indeed. I considered another approach but that's a discussion that we need to have in a different context. Right now this change is just cosmetics.

Ozaq:
Addressed comment by @simondsmart to subclass `eckit::Exception`

PR #89, opened 2025-02-19
FieldLocation shared ptr

mcakircali:
FieldLocationFactory makes shared_ptr
remove unnecessary make_shared method 
use `std::make_shared` and `shared_from_this` as needed
adds test


mcakircali:
too old

PR #91, opened 2025-02-21
FDB-417: Lazily instantiate FDB objects

simondsmart:
By lazily constructing FDB objects, we avoid FDB Remote instances connecting to their counterparts even if those lanes are not going to be used.

This does have two negative consequences:

1. By delaying when the connections happy, we may increase end-to-end times
2. We defer when the config for the sub-FDBs are read, which means we can end up with runtime errors later in the process

Comments welcome.

### DIFF
--- a/src/fdb5/api/SelectFDB.cc
+++ b/src/fdb5/api/SelectFDB.cc
@@ -63,6 +63,19 @@ std::map<std::string, eckit::Regex> parseFDBSelect(const eckit::LocalConfigurati
 }
 
 
+FDB& SelectFDB::FDBLane::get() {
+    if (!fdb) {
+        fdb.emplace(config);
+    }
+    return *fdb;
+}
+
+void SelectFDB::FDBLane::flush() {
+    if (fdb) {
+        fdb->flush();
+    }
+}
+
 SelectFDB::SelectFDB(const Config& config, const std::string& name) : FDBBase(config, name) {
 
     ASSERT(config.getString("type", "") == "select");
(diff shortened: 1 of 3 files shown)


src/fdb5/api/SelectFDB.cc
```diff
@@ -72,7 +85,7 @@ SelectFDB::SelectFDB(const Config& config, const std::string& name) : FDBBase(co
     }
 
     for (const auto& c : config.getSubConfigs("fdbs")) {
-        subFdbs_.emplace_back(std::make_pair(parseFDBSelect(c), FDB(c)));
+        subFdbs_.emplace_back(FDBLane{parseFDBSelect(c), c, std::optional<FDB>{}});
```
Ozaq, line 88: I personally prefer to use `std::nullopt` here instead of `std::optional<FDB>{}}` as I think this is easier to parse for the reader.

src/fdb5/api/SelectFDB.h
```diff
@@ -83,11 +92,10 @@ class SelectFDB : public FDBBase {
 
 private:  // members
 
-    std::vector<std::pair<SelectMap, FDB>> subFdbs_;
+    std::vector<FDBLane> subFdbs_;
```
Ozaq, line 95: Since the code converges on using the term 'lane' I would also rename this field to `lanes_`

Ozaq (commented):
I think it should be added to the API doc of this class that configuration errors can now happen on any call due to the lazy instantiation.

PR #94, opened 2025-03-06
 Add regeression test for FDB-419

Ozaq:
This PR adds a regression test for FDB-419. This test correctly fails when cherry picked to a commit prior of the 'fix/FDB-419' merge.

---
EDIT: ~Back to draft until the liibmf.so issue on intel runner is resolved.~
The libimf.so issue is resolved, it was an issue in the test itself. When fork+execing `fdb-server` I initially overwrote the environment but for Intel libimf.so location is provided via LD_PRELOAD by module load on our runners. Now the existing environment is copied and amended to run `fdb-server`.


ChrisspyB:
>but for Intel libimf.so location is provided via LD_PRELOAD
D:

src/fdb5/remote/client/ClientConnectionRouter.cc
```diff
@@ -83,11 +83,26 @@ std::shared_ptr<ClientConnection> ClientConnectionRouter::connection(
...
 }
 
+std::shared_ptr<ClientConnection> ClientConnectionRouter::refresh(const std::shared_ptr<ClientConnection>& connection) {
+    std::lock_guard lock(connectionMutex_);
+    const auto iter = connections_.find(connection->controlEndpoint());
+    if (iter == connections_.end() || !iter->second->valid()) {
+        auto newConnection =
+            std::make_shared<ClientConnection>(connection->controlEndpoint(), connection->defaultEndpoint());
+        if (newConnection->connect()) {
+            const auto it = (connections_.emplace(newConnection->controlEndpoint(), newConnection)).first;
```
ChrisspyB, line 93: I assume you don't need the `const auto it = ` part anymore? Just emplace the newConnection

tests/regressions/FDB-419/schema
```diff
@@ -0,0 +1,590 @@
+
```
ChrisspyB, line 1: I know a lot of tests don't do this, but I always try to use a reduced schema.

Probably this would be sufficient:

```
param:      Param;
step:       Step;
date:       Date;
levelist:   Double;
expver:     Expver;
time:       Time;

[ class, expver, stream, date, time, domain?
       [ type, levtype
               [ step, levelist?, param ]]
]
```

ChrisspyB (approved):
Minor comments, otherwise looks good.

In a later PR, we should factor out some of this server spawning code to some place where it could be used for other remote FDB tests. Would be nice to have even a very simple test of all the remoteFDB api (though will need to spawn multiple servers for catlaogue+store).

tests/regressions/FDB-419/schema
Ozaq: Will do

PR #95, opened 2025-03-06
fdb-425 unable to list after purging with sub tocs

ChrisspyB:
<s>WIP, not finished.</s> Ready for review.

We are unable to list an FDB where the following has happened:
- Write some fields to an FDB with parallel processes, with subtocs enabled.
- Repeat these writes a second time to mask the data.
- Run an FDB purge, which will TOC_CLEAR the masked indexes and delete them. Importantly, the TOC_CLEAR entry ends up in the parent toc, not the sub toc containing the Index entry.
- ❌ Try to list the FDB: fails as it tries to open non-existing index files.

Listing fails because the TocHandler does not consider the fact that a TOC_CLEAR entry in a parent toc file might be responsible for clearing an entry in a subtoc file.

I have also noticed fdb-purge not clearing subtocs which have no more valid entries. And indeed, there are all manner of things wrong with our treatment of masking when subtocs in in play.


ChrisspyB:
Currently this branch has a test to reproduce the issue, and some rough code to circumvent it. But it is not well integrated with the existing async preloading of the subtocs, which I think will need to be updated.

ChrisspyB:
Hmm, it seems to hang on opening a subtoc on the unix machines (and I've verified it also happens on atos). Investigating...

ChrisspyB:
Removed the forking from the second test for now. No need for that to block this PR.

PR #96, opened 2025-03-11
FDB-427 Limit memory usage on remotefdb.read

ChrisspyB:
(Only relevant to RemoteFDB)

Introduces the ReadLimiter class, which tracks the total size of fields already requested from the store servers and prevents requesting more fields until the exisitng ones have been consumed. The FDBRemoteDataHandle is responsible for signalling to the ReadLimiter when its field has been consumed.

RemoteStores no longer directly send their request to a server, they now first send it to the ReadLimiter class, which holds onto the request until there is sufficient memory to process it.


src/fdb5/remote/RemoteFieldLocation.cc
```diff
@@ -93,7 +93,8 @@ eckit::DataHandle* RemoteFieldLocation::dataHandle() const {
         remote.query("internalHost", "");
     }
     remote.query("internalScheme", "");
-    FieldLocation* loc = FieldLocationFactory::instance().build(scheme, remote, offset_, length_, remapKey_);
+    std::unique_ptr<FieldLocation> loc(
```
ChrisspyB, line 96: NB: this was a fly-by fix of a memory-leak.

src/fdb5/remote/client/ReadLimiter.h
```diff
@@ -0,0 +1,90 @@
...
+    // Attempt to send the next request in the queue. Returns true if a request was sent.
+    // If not enough memory is available, or there is no next request, returns false.
+    bool tryNextRequest();
+
+    void finishRequest(uint32_t clientID, uint32_t requestID);
+
+    // When a RemoteStore is destroyed, it must evict any unconsumed requests.
+    // If all went well, there will be no requests to evict, but we must check (consumer may have abandoned the
+    // request).
+    /// @todo: This is somewhat pointless right now because the RemoteStores appear to be infinitely long lived...
```
simondsmart, line 59: This seems surprising. Are these not owned by something with a finite lifetime?

src/fdb5/remote/client/ReadLimiter.h
```diff
@@ -0,0 +1,90 @@
...
+
+    // Debugging
+    void print(std::ostream& out) const;
+
+private:
+
+    ReadLimiter();
+
+    // Send the request to the server
+    void _sendRequest(const RequestInfo& request) const;
```
simondsmart, line 71: Please don't introduce a new naming convention for private or protected methods. This is not python.

src/fdb5/remote/client/ReadLimiter.h
```diff
@@ -0,0 +1,90 @@
...
+
+    static ReadLimiter& instance();
+
+    // Add a new request to the queue of requests to be sent. Will not be sent until we know we have buffer space.
+    void add(RemoteStore* client, uint32_t id, const FieldLocation& fieldLocation,
+             const Key& remapKey);  // use const *?
+
+    // Attempt to send the next request in the queue. Returns true if a request was sent.
+    // If not enough memory is available, or there is no next request, returns false.
+    bool tryNextRequest();
```
simondsmart, line 52: It is a bit odd that the return value for this method doesn't seem to ever be used

simondsmart (commented):
Some simple comments added.

But overall, I don't understand why this is a thing. We have (size limited) queues for received messages, and for data waiting to be read. Why is that not catching the right stuff?

src/fdb5/remote/client/ReadLimiter.h
ChrisspyB: I figured there could one day be contexts where it was useful to know whether the request was actually sent. But can make void if this is premature.

ChrisspyB:
We would like to have more fine grained control of queue sizes. We have decided this will be implemented in a followup PR, for the sake of expediency. I will just make some small changes here first before we merge.

PR #97, opened 2025-03-14
Use muti-byte string literals for buffer sizes

Ozaq:


ChrisspyB:
This should presumably also update the minimum eckit version in the cmake lists

Ozaq:
> This should presumably also update the minimum eckit version in the cmake lists

Oh good catch, so that should then probably be the next to-be-released version of eckit, right?

PR #98, opened 2025-03-14
Tidy up

Ozaq:
This PR enables a few more tidy checks and addresses them. Additionally it adds a way for git blame to ignore changed from our big reformatting.

[Removed unneeded if](https://github.com/ecmwf/fdb/commit/faa7ac46519754c01fb32b92ec1acc4b13be9b85) 
There is no need to null check a pointer before deleting because 'delete
ptr' where prt = nullptr is a NO-OP.

[Add .git-blame-ignore-revs](https://github.com/ecmwf/fdb/commit/9778fadce2bd58a8e5999af396560c4860b8c0c6) 
'git blame' can be configured to ignore specific commits such as
reformatting the codebase with clang format.

We track the commits we want to be ignored in .git-blame-ignore-revs.

You can use git blame from the cli with:
    git blame --ignore-revs-file .git-blame-ignore-revs <FILE>

Or you configure your git config to always use .git-blame-ignore-revs by
setting:
    git config blame.ignoreRevsFile .git-blame-ignore-revs

Or extending your gitconfig with:

[blame]
    ignoreRevsFile = .git-blame-ignore-revs

Note that .git-blame-ignore-revs is recognized by GitHub.

[Introduce tidy bugprone-use-after-move](https://github.com/ecmwf/fdb/commit/f7b85bea937cf943ebb8a8e6edf6959f3febbfc4) 
Move- constructor / assignment operator can be defaulted in this case.
Silencing the use-after-move warning. In this specific case there was no
bug in the existing code because we moved only the base class part.

[Add aditional tidy checks](https://github.com/ecmwf/fdb/commit/c98cd9eec862bbd4b678917b086330ddd3bb3c95) 
- bugprone-assignment-in-if-condition
- bugprone-bool-pointer-implicit-conversion
- bugprone-unused-return-value,

[Added new tidy-check and fixed warning](https://github.com/ecmwf/fdb/commit/3127dade71aaf5d3375246430571def9b1e169a9) 
Added bugprone-casting-through-void

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,6 +3,7 @@ UseColor: On
 Checks: >-
   -*,
   readability-delete-null-pointer,
+  bugprone-use-after-move,
   bugprone-parent-virtual-call,
 
 WarningsAsErrors:             false
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -5,6 +5,7 @@ Checks: >-
   readability-delete-null-pointer,
   bugprone-assignment-in-if-condition,
   bugprone-bool-pointer-implicit-conversion,
+  bugprone-casting-through-void,
   bugprone-parent-virtual-call,
   bugprone-unused-return-value,
   bugprone-use-after-move,
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,8 +3,12 @@ UseColor: On
 Checks: >-
   -*,
   readability-delete-null-pointer,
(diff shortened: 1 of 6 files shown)


Ozaq:
@simondsmart I talked to @wdeconinck and @pmaciel about https://clang.llvm.org/extra/clang-tidy/checks/readability/delete-null-pointer.html and the resonance was positive.

PR #99, opened 2025-03-17
Bump TocSerialisationVersion to 3

ChrisspyB:
Bump from serialisation version of the toc 2 -> 3 (now includes timestamps), which was falsely assumed to already be the case


Ozaq (approved):
😁 I concede defeat!

PR #102, opened 2025-03-21
FDB-438 extracted code for fdb compact listing from fdb-list tool

danovaro:


src/fdb5/api/FDB.cc
```diff
@@ -330,4 +332,76 @@ void FDB::registerFlushCallback(FlushCallback callback) {
 
 //----------------------------------------------------------------------------------------------------------------------
 
+namespace helper {
+
+static std::string keySignature(const fdb5::Key& key) {
```
ChrisspyB, line 337: In my view it would be better if all this stuff could go to its own file e.g. in fdb5/api/helpers/<new_file>, rather than in FDB.cc.

src/fdb5/api/FDB.h
```diff
@@ -299,6 +299,11 @@ class FDB {
 };
 
 //----------------------------------------------------------------------------------------------------------------------
+namespace helper {
+
+void compact(ListIterator& list, std::ostream& out);
```
ChrisspyB, line 304: Same as other fdb5/api/helpers/<new_file> comment.
Could add an include to the top of FDB.h if you want it to always be included.

ChrisspyB (commented):
As far as I can tell, this is just a cut and paste from the tool 👍 , though I don't think it really belongs in FDB.cc/.h either.

src/fdb5/api/FDB.cc
danovaro: thanks, @ChrisspyB for your suggestion.
Instead of creating a new file, I've added it as a ListIterator method

PR #107, opened 2025-04-14
Feature/zfi/add new library

Ozaq:
@simondsmart please have a look, we encountered the issue that so far we hav no support for tracking multiple build artifacts with diverging version numbers so I had to add some cmake code to facilitate that.


tbkr:
Looks good to me. Can be merged into zfi branch, as soon as the ci passed.

PR #108, opened 2025-04-15
Enable dummy DAOS tests

nicolau-manubens:
🌈🌦️📖🚧 Documentation 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-108


mcakircali (changes requested):
ci configs look good.

the builds fail because the FindUUID.cmake needs fixing such that it finds uuid/libuuid at default locations (on each platform).

simondsmart (approved):

PR #109, opened 2025-04-15
Update all actions with org change

Ozaq:
ecmwf-actions org has been deprecated in favour of just ecmwf. No repo in ecmwf-actions is receiving updates any longer.


ChrisspyB:
ty

PR #116, opened 2025-04-23
FDB-471: Add mutex to TypesRegistry cache

ChrisspyB:
Adds a mutex to the TypesRegistry's internal cache, as a data race occurs when calling `fdb.list` on a SelectFDB.

**HOWEVER** it is unclear to me why multiple threads would be accessing the same TypesRegistry object in the first place. Someone with understanding should look at this, as it seems likely to me that the `types_` member likely ought to be protected by a mutex also if this object is intentionally used in a multithreaded context.


mcakircali (commented):
a unit test would be helpful. so, I would speculate that the race condition was related to `parent_`
maybe needs a check on `parent_` for preventing deadlock but difficult without a test

PR #118, opened 2025-04-29
FDB-474: Fix deadock on disconnect()

ChrisspyB:
The disconnecting_ flag was being used non-atomically, which lead to a race condition on acquiring the clientsMutex_. Since the disconnect() method hangs until all other threads are joined, this led to a deadlock.

I have removed the disconnecting_ flag as I believe it to be redundant, due to the ASSERT(clients_.empty()). I also release the clientsMutex_ lock early in disconnect(), as it does not access the clients_ map after that point. (Additionally, since disconnect() is inside the ClientConnection destructor, something must have gone seriously wrong if clients_ are still mutating at this stage...)

----
📚 Documentation preview 📚: https://ecmwf-fdb--118.org.readthedocs.build/en/118/

### DIFF
--- a/src/fdb5/remote/client/ClientConnection.h
+++ b/src/fdb5/remote/client/ClientConnection.h
@@ -93,8 +93,6 @@ private:  // members
     eckit::net::TCPClient controlClient_;
     eckit::net::TCPClient dataClient_;
 
-    bool disconnecting_ = false;
-
     std::mutex clientsMutex_;
     std::map<uint32_t, Client*> clients_;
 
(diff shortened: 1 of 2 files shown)


danovaro (commented):
now that disconnecting_ is not used anymore, can you please remove from the header?

src/fdb5/remote/client/ClientConnection.h
```diff
@@ -107,8 +107,6 @@ class ClientConnection : protected Connection {
     std::mutex idMutex_;
     uint32_t id_;
 
-    bool connected_;
```
danovaro, line 110: I believe connected_ is still required.
Can you please revert and remove disconnecting_?
tks!

danovaro (approved):
lgtm

PR #127, opened 2025-05-14
FDB-486: Support list-style input in fdb-copy

simondsmart:


nicolau-manubens:
Looks good to me. Seems like FDB-243.sh needs updated.

nicolau-manubens:
I've tested this and it works and performs well.

nicolau-manubens (approved):

src/fdb5/tools/fdb-copy.cc
```diff
@@ -78,43 +128,37 @@ static std::vector<metkit::mars::MarsRequest> readRequest(const CmdArgs& args) {
...
 
-    for (const auto& request : requests) {
-        eckit::Log::info() << request << std::endl;
-        handles.add(fdbRead.retrieve(request));
+    if (fromList_) {
+        for (const FDBToolRequest& request: requests("list")) {
+            bool deduplicate = true;
+            auto listObject = fdbRead.list(request, deduplicate);
+            handles.add(fdbRead.read(listObject, sort_));
+        }
```
jharwood0, line 142: For my own understanding; this isn't using `retreive` for in-line fdb-list partial requests, instead it's doing 2 separate calls, `list` and `read`. Is this to bypass some overhead in `retrieve` where it looks like it's iterating through each field and reading them separately?

nicolau-manubens: The problem is `retrieve` requires a fully specified MARS request to be supplied as input. If we want to copy an entire database, we need to obtain and provide the potentially multiple MARS sub-requests that cover the entire database domain. And `retrieve` will internally `inspect` the FDB indices for each sub-request to extract the field locations, resulting in repeated access and loading of many index files. With `list` we can just specify an incomplete or partial MARS request and it extracts all field locations in one go, extracting all relevant content from index files in one go. Both methods call `read` eventually to wrap the found locations as DataHandles for actual data reading, but this is done all locally on the client side.

PR #132, opened 2025-05-22
Fix fdb.read when using single connection

ChrisspyB:
----
📚 Documentation preview 📚: https://ecmwf-fdb--132.com.readthedocs.build/en/132/

### DIFF
--- a/src/fdb5/remote/Connection.cc
+++ b/src/fdb5/remote/Connection.cc
@@ -130,7 +130,7 @@ void Connection::write(const Message msg, const bool control, const uint32_t cli
 
     MessageHeader message{msg, control, clientID, requestID, payloadLength};
 
-    const auto& socket = control ? controlSocket() : dataSocket();
+    const auto& socket = (control || single_) ? controlSocket() : dataSocket();
     LOG_DEBUG_LIB(LibFdb5) << "Connection::write [endpoint=" << socket.remoteHost() << ":" << socket.remotePort()
                            << ",message=" << msg << ",clientID=" << message.clientID() << ",control=" << control
                            << ",requestID=" << requestID << ",payloadsSize=" << payloads.size()
--- a/src/fdb5/remote/Connection.cc
+++ b/src/fdb5/remote/Connection.cc
@@ -43,13 +43,8 @@ void Connection::teardown() {
 //----------------------------------------------------------------------------------------------------------------------
 
 void Connection::writeUnsafe(const bool control, const void* const data, const size_t length) const {
-    long written = 0;
-    if (control || single_) {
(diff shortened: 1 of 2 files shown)


Ozaq:
How about a regression test?

ChrisspyB:
> How about a regression test?

In a followup PR I think, due to the faff of remoteFDB tests.

I have a number of scripts that I use for testing outside the CI, I will experiment with adding them to the ctest/CI tomorrow (or, do a C++ only approach based on your FDB-419 regression test. Though I think I can do more with cmake + bash).

src/fdb5/remote/Connection.cc
```diff
@@ -43,13 +43,8 @@ void Connection::teardown() {
...
 void Connection::writeUnsafe(const bool control, const void* const data, const size_t length) const {
-    long written = 0;
-    if (control || single_) {
-        written = controlSocket().write(data, length);
-    }
-    else {
-        written = dataSocket().write(data, length);
-    }
+    const auto& socket = getSocket(control);
+    long written       = socket.write(data, length);
```
danovaro, line 47:
```suggestion
  long written = getSocket(control).write(data, length);
```

src/fdb5/remote/Connection.cc
```diff
@@ -65,13 +60,8 @@ void Connection::writeUnsafe(const bool control, const void* const data, const s
...
 bool Connection::readUnsafe(bool control, void* data, size_t length) const {
-    long read = 0;
-    if (control || single_) {
-        read = controlSocket().read(data, length);
-    }
-    else {
-        read = dataSocket().read(data, length);
-    }
+    const auto& socket = getSocket(control);
+    long read          = socket.read(data, length);
```
danovaro, line 64:
```suggestion
    long read = getSocket(control).read(data, length);
```

danovaro (approved):
looks good to me.
I've two small suggestions since we do not use the socket later in the readUnsafe/writeUnsafe methods

src/fdb5/remote/Connection.cc
ChrisspyB: done

src/fdb5/remote/Connection.cc
ChrisspyB: done

src/fdb5/remote/Connection.h
```diff
@@ -79,7 +79,13 @@ class Connection : eckit::NonCopyable {
 
     void writeUnsafe(bool control, const void* data, size_t length) const;
```
danovaro, line 80: do we need to keep it?

PR #133, opened 2025-05-27
FDB-505: Add remoteFDB api ctest and test FDB-491

ChrisspyB:
- Add ctest that spawns a catalogue and store server in the background, and does a simple write/list/read, using a "standard" set of configs.
- Add a second test which is the same as above, but with a config change (using single connection. This is a regression test for issue FDB-49.

The underlying test is reusable and could be reused to test other sets of configs if need be.

Draft for now as I want to see how it behaves on the CI. *Should* work fine...

----
📚 Documentation preview 📚: https://ecmwf-fdb--133.com.readthedocs.build/en/133/

🌈🌦️📖🚧 Documentation 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-133


danovaro (approved):
looks good,
only warning, on MacOS timeout is missing
I fixed with
brew install coreutils

PR #135, opened 2025-06-10
Integrate pyfdb

Ozaq:
----
📚 Documentation preview 📚: https://ecmwf-fdb--135.com.readthedocs.build/en/135/


ChrisspyB:
It might be good to do this in a way that preserves history e.g. using git subtree.
I did something like this when merging pymetkit into metkit.

simondsmart:
If there is a way of doing that, please do so. Otherwise (assuming this is a straight transfer of functionality) then it should just be merged.

PR #145, opened 2025-06-24
FDB-518: Use a 3-level map/set of key fingerprints to disambiguate fdb list output

simondsmart:
🌈🌦️📖🚧 Documentation 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-145

### DIFF
--- a/src/fdb5/api/helpers/ListIterator.h
+++ b/src/fdb5/api/helpers/ListIterator.h
@@ -50,9 +50,7 @@ using ListAsyncIterator = APIAsyncIterator<ListElement>;
 
 class KeyStore {
 
-    std::unordered_map<std::string,
-        std::unordered_map<std::string,
-            std::unordered_set<std::string>>> fingerprints_;
+    std::unordered_map<std::string, std::unordered_map<std::string, std::unordered_set<std::string>>> fingerprints_;
 
 public:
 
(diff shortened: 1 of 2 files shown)


src/fdb5/api/helpers/ListIterator.h
```diff
@@ -48,6 +48,19 @@ using ListAsyncIterator = APIAsyncIterator<ListElement>;
 
 //----------------------------------------------------------------------------------------------------------------------
 
+class KeyStore {
+
+    std::unordered_map<std::string,
+        std::unordered_map<std::string,
+            std::unordered_set<std::string>>> fingerprints_;
```
Ozaq, line 55: I think instead of having 2 indirections in memory you could also just use a single `std::unordered_set<std::array<Key, 3>>` with a custom hash function.

Edit: Hmm after looking at how the hash for Key we probably should use an incremental hasher from xxhash (available in eckit), currently we create one big string that then gets hashed which is pretty expensive just for a hash.

If we would use the incremental hasher we could easily create the combined hash for all 3 levels by just streaming in the names.

Ozaq (commented):
Its fine to merge as is, because it is already an improvement BUT there is more potential imo 😃

PR #149, opened 2025-06-27
fdb-copy with modifiers

nicolau-manubens:
### Contributor Declaration

🌈🌦️📖🚧 Documentation 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-149

### DIFF
--- a/src/fdb5/tools/fdb-copy.cc
+++ b/src/fdb5/tools/fdb-copy.cc
@@ -143,7 +143,7 @@ void FDBCopy::checkModifiers(const metkit::mars::MarsRequest& request, const eck
         std::ostringstream msg;
         msg << "Provided modifiers for key '" << pair.first << "' not present in data to be copied";
         if (!request.has(pair.first))
-            throw eckit::UserError(msg);
+            throw eckit::UserError(msg.str());
     }
 }
 
@@ -176,7 +176,6 @@ void FDBCopy::execute(const CmdArgs& args) {
     }
 
     std::unique_ptr<eckit::DataHandle> dh(handles.dataHandle());
-
 
     fdb5::MessageArchiver fdbWriter(fdb5::Key(), false, verbose_, writeConfig);
     fdbWriter.setModifiers(modifiers);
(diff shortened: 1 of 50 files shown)


src/fdb5/tools/fdb-copy.cc
```diff
@@ -56,6 +58,10 @@ FDBCopy::FDBCopy(int argc, char** argv) : fdb5::tools::FDBVisitTool(argc, argv,
     options_.push_back(new SimpleOption<eckit::PathName>("source", "Configuration of FDB to read from"));
     options_.push_back(new eckit::option::SimpleOption<bool>(
         "from-list", "Interpret argument(s) as fdb-list partial requests, rather than request files"));
+    options_.push_back(new eckit::option::SimpleOption<std::string>(
+        "modifiers",
+        "List of comma separated key-values of modifiers to each message "
+        "in input data, e.g --modifiers=packingType=grib_ccsds,expver=0042"));
```
danovaro, line 64: @simondsmart are we happy to allow every kind of grib manipulation supported by ecCodes grib_set?
Should we restrict it to the indexing axes?

simondsmart: I don't think that we should permit arbitrary changes at this time. We could have a discussion about what is permitted later, but I would keep this to the same functionality that we have in the overlay remapping.

That is that we ensure that each key specified is one that is in the original request as well.

Very happy to add an --unsafe argument to turn the verification off, if Nicolau has use cases in mind.

src/fdb5/tools/fdb-copy.cc
```diff
@@ -155,6 +163,7 @@ void FDBCopy::execute(const CmdArgs& args) {
     std::unique_ptr<eckit::DataHandle> dh(handles.dataHandle());
 
     fdb5::MessageArchiver fdbWriter(fdb5::Key(), false, verbose_, writeConfig);
+    fdbWriter.modifiers(modifiers_);
```
simondsmart, line 166: Is there a reason this is done as a separate call? Looks a touch odd.

At least needs a verb.

nicolau-manubens: Now improved.

danovaro (approved):
lgtm

PR #155, opened 2025-07-15
Return hacky extra uri info from c api

TomHodson:
I need a way to get the full location information out. This is probably not the right way to do this, but it is *a* way to do it.
 

🌈🌦️📖🚧 Documentation 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-155


ChrisspyB:
We can do it properly there's no need for it to be hacky

PR #160, opened 2025-08-12
FDB-535; handle index keys >32 chars

jharwood0:
### Contributor Declaration

🌈🌦️📖🚧 Documentation 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-160


ChrisspyB:
I dread to think what issues arose leading to this PR...

src/fdb5/toc/BTreeIndex.cc
```diff
@@ -144,19 +144,29 @@ void TBTreeIndex<KEYSIZE, RECSIZE, PAYLOAD>::preload() {
...
+const std::string& BTreeIndex::defaultType(int version) {
+    static const std::string overrideType = eckit::Resource<std::string>("fdbIndexType;$FDB_INDEX_TYPE", "");
+    if (!overrideType.empty()) {
+        return overrideType;
+    }
+
+    static const std::string defaultType   = "BTreeIndex";
+    static const std::string defaultTypeV4 = "BTreeIndexV4";
+
+    return (version == 4) ? defaultTypeV4 : defaultType;
```
simondsmart, line 164: Note that we don't want to do this.

The effect of this is that data written by new clients will immediately become unreadable to older clients.

We want to default to the same behaviour, and only shift to using the updated type if needed. Whether that means explicitly specified, or by some heuristic based on the number of keys in the rule, is to be determined.

I'm not convinced that we need to update the serialisation version. We are adding a value, but I don't think we change the format per se.

simondsmart: No, lets be careful. This keeps the default serialisation version to 3. But if the serialisation version changes, the default will change. I'm not sure we want to change that default.

And, for that matter, it isn't obvious to me that we need to change the serialisation version. That is going to cause all old clients to bail out. And we don't want that.

PR #165, opened 2025-08-26
Feature/fdb 327 fdb compare

stefaniereuter:
### Description
Opening this as a draft pull request, while finishing the refactoring on the grib comparison but to start discussing the design and features. 

This draft shows the direction in which I would like to continue refactoring. 

This tool allows to compare two FDB or two experiments in one FDB. It uses the standard fdb tool mechanisms, minimal-keys, CLI request, but also allows to directly specify mars keys, where entries should be ignored. it uses fdb list to generate a request then sorts the results and compares them, depending on scope, where the default is just a comparison of the mars keys, without the messages itself, possibility of comparing grib-header only and then grib-header plus data section. Default is a key by key comparison, 
More explanation and examples can be found here:

https://confluence.ecmwf.int/display/~ecm4053/FDB+Comparison

### Contributor Declaration


tests/fdb/tools/compare/mismatch_grib/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+ecbuild_configure_file( mismatch_grib.sh.in mismatch_grib.sh @ONLY )
+
```
simondsmart: Do the data.grib files differ between the different tests? It looks like you are grib setting them, to set upthe tests on the go. So we probably only need one grib file. You migt even be able to use one which is already referenced in a different test, which would avoid the extra data.

src/fdb5/tools/compare/fdb-compare.cc
```diff
@@ -0,0 +1,294 @@
...
+
+
+class FDBCompare : public FDBVisitTool {
+public:  // methods
+
+    FDBCompare(int argc, char** argv) : FDBVisitTool(argc, argv, "class,expver") {
+        options_.push_back(new SimpleOption<std::string>("test-config", "Path to a FDB config"));
+        options_.push_back(new SimpleOption<std::string>("reference-config", "Path to a second FDB config"));
+        options_.push_back(new SimpleOption<std::string>(
+            "reference-request", "Mars key request for reference FDB entry (e.g reference experiment)"));
```
simondsmart: What is a "mars key request" that is not a phrase that makes sense. At the least this needs an example.

src/fdb5/tools/compare/common/DataMap.h
```diff
@@ -0,0 +1,81 @@
...
+    }
+};
+
+struct MapEqual {
+    bool operator()(const KeyValueMap& a, const KeyValueMap& b) const noexcept { return a == b; }
+};
+
+using DataIndex = std::unordered_map<KeyValueMap, DataLocation, MapHash, MapEqual>;
+
+/// Print a KeyValueMap
```
simondsmart: Unclear why these warrant a comment, when they are absolutely clear what they do, but there is no commentary/docs about what the purpose of these classes is.

src/fdb5/tools/compare/mars/Compare.cc
```diff
@@ -0,0 +1,71 @@
...
+    check(ref, test, "Reference->Test", false);
+    return false;
+}
+
+}  // namespace
+
+
+Result compareMars(const DataIndex& ref, const DataIndex& test, const Options& opts) {
+    Result r;
+    r.match = compareMarsKeys(ref, test, opts.marsReqDiff);
```
simondsmart: Can we not have this Result object either initialised by construction, or being returned directly from compareMarsKeys?

I'm not really clear what value this wrapper adds.

src/fdb5/tools/compare/mars/Compare.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace compare::mars {
+
+/// Compare the mars keys of two messages from different locations.
+///
+/// @param ref  Reference location of the message
+/// @param test Test location of the message
+/// @param opts Additional options - notably marsReqDiff is used to lookup and compare requests
+/// @return Comparison result. As mars is not comparing data, only matched is expected to be set
+Result compareMars(const DataIndex& ref, const DataIndex& test, const Options& opts);
```
simondsmart: This compares the mars keys. Why is it not called ```compareMarsKeys```

src/fdb5/tools/compare/grib/Utils.cc
```diff
@@ -0,0 +1,92 @@
...
+#include <unordered_set>
+
+namespace compare::grib {
+
+
+const std::unordered_set<std::string>& mutableDataKeys() {
+    // Set of keys that are changing if the data values change.
+    // Depending on the packing type different key words are relevant.
+    static const std::unordered_set<std::string> s = {
+        "values",
```
simondsmart: Is there not a canonical location that this information can be obtained from?

src/fdb5/tools/compare/grib/Utils.cc
```diff
@@ -0,0 +1,92 @@
...
+
+    // get first message handle
+    if (refFile.is_open()) {
+        refFile.seekg(gribLoc.offset, std::ios::beg);
+        // For a tco79 this seems to be the bottleneck: Optimisation suggestions: would be to keep the files open in a
+        // Filestream cache. Potentially with an option to only store a certain amount of filestreams to avoid memory of
+        // open file distriptor issues.
+        if (!refFile.read(reinterpret_cast<char*>(buffer.get()), gribLoc.length)) {
+            refFile.close();
+            throw eckit::Abort("Could not read the data from file", Here());
```
simondsmart: Is Abort really the right exception to be throwing here?

src/fdb5/tools/compare/fdb-compare.cc
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    auto marsRes = mars::compareMars(refIdx, testIdx, opts_);
+    if (!marsRes.match) {
+        std::cerr << "[MARS KEYS COMPARE] MISMATCH\n";
+        return;
+    }
+
+
+    // Return if only a comparison of Mars metadata messages was specified as Command Line option
```
simondsmart: Why does _only_ this line warrant a comment. There are bits that could do with explanatory code. Not convinced this is the most significant...

simondsmart (changes requested):

src/fdb5/tools/compare/grib/CompareBitwise.cc
```diff
@@ -0,0 +1,241 @@
...
+        (static_cast<char>(bufferRef[3]) != 'B') || (static_cast<char>(bufferTest[3]) != 'B')) {
+        std::cerr << "[GRIB COMPARISON MISMATCH] Message is not a Grib message: Comparing data formats other than Grib "
+                     "messages is currently not implemented"
+                  << std::endl;
+        return 0;
+    }
+
+    // EditionNumber will is always the 8th Byte in a Grib message
+    int editionnumberRef  = (int)bufferRef[7];
+    int editionnumberTest = (int)bufferRef[7];
```
simondsmart: Both `editionnumberRef` and `editionnumberTest` are read from bufferRef, and so are equal by construction. Should be using `bufferTest`.

Please add a test to check that this is working.

src/fdb5/tools/compare/grib/CompareKeys.cc
```diff
@@ -0,0 +1,217 @@
...
+                  << " CODES_TYPE " << typeRef << "\nTest value: " << valTest << " CODES_TYPE " << typeTest
+                  << std::endl;
+        return {false, false};  // DataSectionMismatch is always false unless the difference is cause by the
+                                // actual values
+    }
+
+
+    if (holdsVector(valRef)) {
+        if (holdsVector(valTest)) {
+            auto l1 = vectorLength(valRef);
```
simondsmart: Note that `vectorLength` returns `bool`. So this is only testing if they are both empty/both have data. Doesn't test lengths.

pgeier: fixed.

tests/fdb/tools/compare/mismatch_mars/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+ecbuild_configure_file( mismatch_mars.sh.in mismatch_mars.sh @ONLY )
+
```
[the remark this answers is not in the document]
pgeier: i'm reusing the data file now

src/fdb5/tools/compare/common/Util.h
```diff
@@ -0,0 +1,71 @@
...
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation nor
+ * does it submit to any jurisdiction.
+ */
+#pragma once
+#include <cstdint>
+#include <cstring>
+#include <iostream>
+
```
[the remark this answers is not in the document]
pgeier: I've added a few separators. 

RE the @author I would suggest relying on git to track this information.

src/fdb5/tools/compare/common/DataMap.h
```diff
@@ -0,0 +1,81 @@
...
+
+namespace compare {
+
+struct DataLocation {
+    std::string path;
+    long long offset = 0;
+    long long length = 0;
+};
+
+using KeyValueMap = std::map<std::string, std::string>;
```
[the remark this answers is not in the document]
pgeier: used std::hash and std::equal_to instead of custom functor names.

I've not changed the naming for now - don't know what a better name would look like. The code is isolated just for comparison.

src/fdb5/tools/compare/mars/Compare.h
pgeier: Renamed.

src/fdb5/tools/compare/grib/Utils.cc
pgeier: AFAIK No

src/fdb5/tools/compare/common/Util.h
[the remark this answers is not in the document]
simondsmart: NO. Git solves a different problem. It tells you who modified each line of code. It does not tell you who is the person/people holding the mental picture of a part of the code.

src/fdb5/tools/compare/common/Types.h
```diff
@@ -0,0 +1,192 @@
...
+//---------------------------------------------------------------------------------------------------------------------
+
+/// Describes the location of a message
+struct DataLocation {
+    std::string path;
+    long long offset = 0;
+    long long length = 0;
+};
+
+/// String-string map which is used to store full FDB keys (i.e. mars requests e.g. {class=rd, expver=1234, ...})
```
simondsmart: I really don't like this comment. Full FDB keys are NOT MARS requests. They don't have a concept of span, and they don't have a verb. This is very misleading.

Secondly, given we are in the FDB repo, I would like to understand why we are not using the fdb5::Key class for representing an FDB key. That would seem like the natural class to use.

src/fdb5/tools/compare/grib/CompareBitwise.cc
```diff
@@ -0,0 +1,249 @@
...
+        // Memory comparison for each section
+        if (std::memcmp(bufferRef + offset, bufferTest + offset, sectionLengthRef) != 0) {
+            std::cout << "[GRIB COMPARISON MISMATCH] Memorycomparison failed in section " << i + 1 << std::endl;
+            return ((i + 1) == numSections) ? 9 : i + 1;
+        }
+        offset += sectionLengthRef;
+    }
+
+    // Test that the final bit of the message 7777 is present in both
+    for (; offset < length; ++offset) {
```
simondsmart: This doesn't test that there are 4 '7' characters. Just that any that there are are '7'.
It also does not test if equal (i.e. it fails just as well if both are identical but wrong)

src/fdb5/tools/compare/common/Types.cc
```diff
@@ -0,0 +1,306 @@
...
+}
+
+
+DataIndex assembleCompareMap(fdb5::FDB& fdb, const fdb5::FDBToolRequest& req, const KeyValueMap& ignore) {
+    DataIndex out;
+
+    auto list = fdb.list(req);
+    fdb5::ListElement elem;
+
+    std::cout << req << std::endl;
```
simondsmart: This looks like debug output that hasn't been removed?

src/fdb5/tools/compare/common/Types.cc
```diff
@@ -0,0 +1,306 @@
...
+
+#include "eckit/exception/Exceptions.h"
+
+#include <sstream>
+
+namespace compare {
+
+//---------------------------------------------------------------------------------------------------------------------
+
+std::ostream& operator<<(std::ostream& os, const KeyValueMap& km) {
```
simondsmart: All of these are completely unneeded if you use the types that are built into the fdb project.

src/fdb5/CMakeLists.txt
```diff
@@ -520,6 +520,25 @@ ecbuild_add_executable(
...
+                                  tools/compare/common/Util.cc
+                                  tools/compare/mars/Compare.cc
+                                  tools/compare/grib/Compare.cc
+                                  tools/compare/grib/CompareKeys.cc
+                                  tools/compare/grib/CompareHash.cc
+                                  tools/compare/grib/CompareDatasection.cc
+                                  tools/compare/grib/CompareBitwise.cc
+                                  tools/compare/grib/Utils.cc
+                        INCLUDES  ${ECCODES_INCLUDE_DIRS}  # Please don't remove me, I am needed
+                                  tools/compare/grib/
```
simondsmart: Why are these needed. This suggests that the include paths have not been properly qualified. Please fix that, and remove these.

src/fdb5/tools/compare/grib/Compare.cc
```diff
@@ -0,0 +1,210 @@
...
+            oss << "Difference in Grib message: Absolute Error = " << res.absoluteError->avg()
+                << " relative Error = " << res.relativeError->avg() << "\n";
+            std::cout << oss.str() << std::endl;
+            res.match = false;
+        }
+    }
+    return res;
+}
+
+void printProgressBar(std::ostream& os, std::size_t current, std::size_t total, std::size_t bar_width = 30) {
```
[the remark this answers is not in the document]
pgeier: I have not seen it anywhere. If you sure we have it, please point me to it.

src/fdb5/tools/compare/common/Util.cc
```diff
@@ -0,0 +1,27 @@
...
+uint32_t readU24Be(const uint8_t* p) {
+    return (uint32_t(p[0]) << 16) | (uint32_t(p[1]) << 8) | uint32_t(p[2]);
+}
+
+uint32_t readU32Be(const uint8_t* p) {
+    return (uint32_t(p[0]) << 24) | (uint32_t(p[1]) << 16) | (uint32_t(p[2]) << 8) | uint32_t(p[3]);
+}
+
+uint64_t readU64Be(const uint8_t* p) {
+    return (uint64_t(p[0]) << 56) | (uint64_t(p[1]) << 48) | (uint64_t(p[2]) << 40) | (uint64_t(p[3]) << 32) |
```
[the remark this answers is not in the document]
simondsmart: That is a fair comment, and I stand corrected on that. They still read a touch oddly to me (I don't really like the magic numbers). Feels like there could be something generic done. Not sure what perhaps something like:

```
std::accumulate(b.begin(), b.end(), 0, [](uint64_t a, uint8_t b) { return (a << 8) | b; });
```

or similar, such that we don't have the sequence of magic numbers. Either way, not as critical.

src/fdb5/CMakeLists.txt
pgeier: removed

src/fdb5/tools/compare/common/Types.h
pgeier: Using fdb5::Key now.

simondsmart (approved):
Looks good. Let's do it.

Ozaq:
Claude Review:
---
# Code Review: `fdb-compare` Tool (commit 75d9ae93)

## Overview

New CLI tool for comparing two FDB (Fields DataBase) instances. Supports three
comparison scopes (MARS keys, header-only, full data) and three GRIB comparison
methods (key-by-key, hash, bit-identical). Well-structured with clear
separation between MARS-level and GRIB-level comparison.

## Architecture & Design

- **Good**: Clean separation into `common/`, `mars/`, and `grib/` modules
- **Good**: Strategy-pattern-like dispatch for comparison methods in
  `gribCompareSingleMessage`
- **Good**: `CompareResult` enum with
  `Match`/`OtherMismatch`/`DataSectionMismatch` allows deferred numeric
  comparison
- **Good**: IIFE pattern for `CompFDB` initialization and method dispatch is
  idiomatic

## Bugs & Correctness Issues

1. **`NumericError::avg()` division by zero** (`Types.cc:186-187`): When `count
   == 0`, `avg()` returns `0.0/0` = NaN. Should guard against this.

2. **GRIB1 total length read as 24-bit** (`CompareBitwise.cc:169`): Uses
   `readU32Be(bufferRef + 4)` but GRIB1 stores total length as 3 bytes at
   offset 4-6 (byte 7 is edition number). Should use `readU24Be`. This
   contradicts the correct usage in `parseGribHeader` at line 54.

3. **`bitComparison` uses `length` from test only** (`Compare.cc:37`):
   `gribLocTest.length()` is passed but `gribLocRef.length()` is never checked.
   If ref and test have different lengths, the comparison may read out of
   bounds or miss the mismatch.

4. **`errCode` logic always returns `OtherMismatch`**
   (`CompareBitwise.cc:195`):

   ```cpp
   const auto errCode = i < numSections ? CompareResult::OtherMismatch : CompareResult::DataSectionMismatch;
   ```

   Since `i` iterates `0..numSections-1`, `i < numSections` is **always true**,
   so `errCode` is always `OtherMismatch`. The intent was likely `i <
   numSections - 1` to distinguish header sections from the data section.

5. **Unused variables** (`CompareKeys.cc:208-209`): `matchreturn` and
   `dataSectionreturn` are declared but never used.

6. **Test script bug** (`all_identical.sh.in:81`): `[ $tool_rc -e 0 ]` should
   be `[ $tool_rc -ne 0 ]`. The `-e` flag tests file existence, not numeric
   equality.

## Performance

1. **`assembleCompareMap` copies entire DataIndex** (`mars/Compare.cc:29`):
   `auto refRemaining(ref)` copies the full map including all `ListElement`s.
   Consider using a `std::unordered_set<fdb5::Key>` of remaining keys instead.

2. **File I/O bottleneck acknowledged** (`Utils.cc:77-79`): The comment about
   keeping file handles open is valid. For large FDBs, opening/closing per
   message will dominate runtime. Worth tracking as a follow-up.

3. **`extractGribMessage` allocates per-message** (`Utils.cc:69-85`): For
   sequential comparisons, reusing a buffer would reduce allocations. Minor for
   now.

## Code Quality

1. **`parseMethod` silently defaults** (`Types.cc:138-141`): Unknown strings
   silently fall through to `KeyByKey`. Should throw on unrecognized input
   (like `parseScope` does) to avoid user confusion.

2. **Typos in output strings**:
    - `Compare.cc:115`: "Differenene" should be "Difference"
    - `CompareKeys.cc:143`: Missing space: `"lengthTest= "` vs `" lengthRef= "`

3. **`fdb-compare.cc:11`**: File header says `fdb-compare.h` but this is a
`.cc` file.

4. **`fdb-compare.cc:240`**: `compare::operator<<(eckit::Log::info(),
opts_.marsReqDiff)` — calling the operator explicitly with qualified name is
unusual. Using `eckit::Log::info() << opts_.marsReqDiff` would be cleaner if
ADL finds it (may need a `using` declaration).

## Security

1. **Buffer reads in `bitComparison`**: The `ASSERT(length > 16)` at line 133
is good, but the section-walking loop at lines 189-229 trusts
`sectionLengthRef` from the buffer without bounds-checking against the total
`length`. A malformed GRIB could cause out-of-bounds reads via `offset +
sectionLengthRef > length`.

## Test Coverage

- **Good**: Four test scenarios covering identical FDBs, GRIB header
  mismatches, data mismatches with tolerance, and different experiment versions
- **Missing**: No unit tests for individual functions (`compareValues`,
  `bitComparison`, `requestDiff`, etc.)
- **Missing**: No test for error paths (corrupt GRIB, missing config files,
  empty FDBs)
- **Missing**: No test for the `hash` comparison method
- **Bug in test**: `all_identical.sh.in:81` uses `-e` instead of `-ne` (see
  item 6)

## Commit Message

The commit message "Added fdb-compare tool to compare two FDBs" is inadequate
for a 2760-line, 46-file addition. It should:

- Explain the motivation (why is this tool needed, what workflow does it
  enable?)
- Summarize the architecture (three comparison scopes, three GRIB methods, MARS
  key diffing)
- Reference a ticket/issue number if one exists
- Mention key design decisions (e.g. why multiple comparison strategies,
  tolerance-based FP comparison)

## End-User Documentation

`docs/fdb/cli_tools/compare.rst` exists and covers options reasonably, but has
issues:

- **Typos and formatting**: "comparsion" (line 7), unclosed backtick in example
  on line 79 (`config.yaml` missing closing `>` and has trailing \` ),
  unbalanced parentheses on lines 68 and 70
- **Example uses wrong binary name** (line 84): `fdbcompare` should be
  `fdb-compare`
- **No explanation of exit codes**: The tool returns 0 on match and 1 on
  mismatch (used extensively in tests), but this is undocumented. Critical for
  scripted usage.
- **No description of output format**: Users need to know what `[MARS KEYS
  COMPARE] SUCCESS`, `[GRIB COMPARISON MISMATCH]`, etc. mean and how to parse
  them
- **Missing guidance on method selection**: No advice on when to use
  `grib-keys` vs `hash-keys` vs `bit-identical` (trade-offs between speed,
  granularity of error reporting, and strictness)
- **No explanation of `--mars-keys-ignore` vs `--grib-keys-ignore`**: The
  difference between filtering at MARS level (excluding entire entries) vs GRIB
  level (skipping specific metadata keys) is not clarified
- **Single-FDB mode undocumented**: Using `--config` alone (without
  `--reference-config`/`--test-config`) to compare within one FDB is only
  discoverable by reading the source

## API Documentation (Headers)

Most public headers have doxygen `@param`/`@return` comments, which is
positive. Issues:

- **`CompareBitwise.h:26`**: `compareHeader()` has no documentation at all — no
  description, no `@param`, no `@return`
- **`Utils.h:34-36`**: `isDataKey` doc is oddly structured — `@return` says
  "Check if a given key is a data key" but that's a description, not a return
  value. Should be `@return true if the key is data-relevant`
- **`Utils.h:39-40`**: `appendDataRelevantKeys` has `@return` but returns
  `void`
- **`CompareResult` enum** (`Utils.h:28-32`): No documentation on the enum
  values. `DataSectionMismatch` vs `OtherMismatch` semantics are non-obvious
  and important for callers
- **`Types.h`**: `Options` struct fields have inline comments (good), but
  `NumericError` and `Result` have no member-level docs. `NumericError::min`
  being initialized to `0.0` rather than `+inf` is surprising and undocumented
  — first call to `update()` will never increase `min` above 0 even if all
  errors are positive
- **`mars/Compare.h:19`**: `using namespace compare;` in a header pollutes the
  namespace of all includers

## Summary

The tool is well-architected with a clean module layout. The main concerns are:

- **Critical**: GRIB1 length parsing mismatch (item 2), buffer bounds issue
  (item 14), test bug (item 6)
- **Medium**: Division-by-zero in `avg()` (item 1), `errCode` always
  `OtherMismatch` (item 4), silent default for unknown methods (item 10),
  commit message quality, documentation gaps
- **Low**: Unused variables, typos, performance notes, API doc inconsistencies

Ozaq:
### Code review (Claude)

Found 6 issues:

1. **`fdb-compare` not guarded by `HAVE_GRIB` in CMakeLists.txt** -- `fdb-compare` depends on eccodes/GRIB APIs but is only gated by `HAVE_FDB_BUILD_TOOLS`, not `HAVE_GRIB`. Other GRIB-dependent tools (`fdb-hammer`, `fdb-patch`) are correctly guarded with `if ( HAVE_GRIB )`. Building with `HAVE_GRIB=OFF` will fail.

https://github.com/ecmwf/fdb/blob/ea44ead64db6ab20657c408c6b56d7fbc55642e2/src/fdb5/CMakeLists.txt#L522-L546

2. **Data section tolerance check uses `avg()` instead of `max`** -- The comparison decides match/mismatch based on `avg() > tolerance`. A few values with errors 100x the tolerance will be averaged away if most values are close, causing the tool to silently accept GRIB messages with large localized errors. The `max` field is already tracked in `NumericError` but never consulted in the decision logic.

https://github.com/ecmwf/fdb/blob/ea44ead64db6ab20657c408c6b56d7fbc55642e2/src/fdb5/tools/compare/grib/Compare.cc#L119-L127

3. **`DataHandle` not closed after read in `extractGribMessage`** -- `dh->openForRead()` is called but `dh->close()` is never called. The `unique_ptr` destructor destroys the handle but `eckit::DataHandle::~DataHandle()` does not call `close()`. This leaks a file descriptor per GRIB message compared. The established eckit pattern is to use `eckit::AutoClose`.

https://github.com/ecmwf/fdb/blob/ea44ead64db6ab20657c408c6b56d7fbc55642e2/src/fdb5/tools/compare/grib/Utils.cc#L74-L83

4. **`requests()[0]` and `requestsFromString(*s)[0]` without bounds check** -- Both access `[0]` on vectors that may be empty (e.g., user provides no positional arguments, or malformed `--reference-request`). This is undefined behaviour.

https://github.com/ecmwf/fdb/blob/ea44ead64db6ab20657c408c6b56d7fbc55642e2/src/fdb5/tools/compare/fdb-compare.cc#L226-L231

5. **`--fp-tolerance` help text claims default is "machine tolerance epsilon" but actual default is `0.0`** -- `Options::tolerance` is initialized to `0.0` in `Types.h:132`, not machine epsilon. Users who rely on the help text will have incorrect expectations about default behavior.

https://github.com/ecmwf/fdb/blob/ea44ead64db6ab20657c408c6b56d7fbc55642e2/src/fdb5/tools/compare/fdb-compare.cc#L75-L76

6. **RST title underline too short in `compare.rst`** -- "fdb-compare" is 11 characters but the underline is only 9 `=` characters. This will cause a Sphinx build warning or error.

https://github.com/ecmwf/fdb/blob/ea44ead64db6ab20657c408c6b56d7fbc55642e2/docs/fdb/cli_tools/compare.rst#L1-L2

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #246, opened 2026-03-07
schema - optional with defaults - improved behaviour (not injecting the default in the index)

danovaro:
…e index

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/dev-section/fdb/pull-requests/PR-246


src/fdb5/CMakeLists.txt
```diff
@@ -475,7 +475,7 @@ list( APPEND fdb5_tools
 if ( HAVE_GRIB )
     list( APPEND fdb5_tools
         fdb-hammer
-        fdb-patch
+        # fdb-patch
     )
```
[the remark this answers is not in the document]
danovaro: protected with a conditional

src/fdb5/api/FDB.cc
```diff
@@ -237,7 +241,19 @@ eckit::DataHandle* FDB::retrieve(const metkit::mars::MarsRequest& request) {
...
 ListIterator FDB::inspect(const metkit::mars::MarsRequest& request) {
-    return internal_->inspect(request);
+
+    ASSERT(!request.empty());
+
+    metkit::mars::MarsRequest indexingRequest{request.verb()};
+    metkit::mars::MarsLanguage language{request.verb()};
+    // Only copy data parameters, as these are the only ones relevant for the indexing
+    for (const auto& p : request.parameters()) {
+        if (!language.isPostProc(p.name()) && !language.isSink(p.name())) {
```
simondsmart: I don't like this. Why is this ever happening? And why is this change a part of a PR related to ?-type optional behaviour.

These keywords should never make it here.

src/fdb5/database/IndexAxis.cc
```diff
@@ -399,10 +399,13 @@ void IndexAxis::print(std::ostream& out) const {
 
     const char* sep = "";
     out << "{";
-    for (const auto& kv : axis_) {
-        out << sep << kv.first << "=(";
+    for (const auto& [k, vv] : axis_) {
+        if (vv->size() == 1 && vv->at(0).empty()) {
```
simondsmart, line 403: If vv->size() != 1, but one of the values is empty, what is the correct behaviour here?

src/fdb5/rules/Matcher.h
```diff
@@ -47,6 +47,11 @@ class Matcher : public eckit::Streamable {
     virtual const std::string& value(const Key&, const std::string& keyword) const;
     virtual const std::vector<std::string>& values(const metkit::mars::MarsRequest& rq,
                                                    const std::string& keyword) const;
+
+    /// Returns by value to allow subclasses to construct custom lists
```
simondsmart: I don't think that suggests this should return by value. It suggests that we should memoise the result (i.e. store the result somewhere in the class, constructing it on first call to optionalValues()). Otherwise we are potentially doing a lot of repetitive, entirely unnecessary memory manipulation.

simondsmart (approved):
Looks good. Just two quick queries:

- copilot review notes a discrepancy between fdb-patch in two places in the codebase. Has this been checked.
- I really don't understand why we are changing the behaviour of inspect() wrt. pproc and sink parameters. And certainly not why that is happening in this PR. I don't think that change is conceptually correct - if those parameters are present, they belong to a tool higher up in the chain, and should be being removed there not in the FDB.

src/fdb5/api/FDB.cc
danovaro: thanks, removed.

src/fdb5/rules/Matcher.h
danovaro: modified to return a const ref, memoised in MatchOptional

src/fdb5/database/IndexAxis.cc
danovaro: vv is sorted every time we decode the Axes.
if vv->size() > 1, only the first value can be empty (due to the ordering).
In that case, we print `key=(,v2,v3)`

PR #267, opened 2026-04-15
add control flags to fdb remote networks

danovaro:
### Description

add ControlIdentifier to the remote FDB networs (to prevent specific operations from selected networks) 

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-267

### DIFF
--- a/src/fdb5/api/helpers/ControlIterator.cc
+++ b/src/fdb5/api/helpers/ControlIterator.cc
@@ -86,6 +86,24 @@ ControlIdentifiers::ControlIdentifiers(eckit::Stream& s) {
     s >> value_;
 }
 
+ControlIdentifiers::ControlIdentifiers(const eckit::LocalConfiguration& config) : value_(0) {
+
+    bool writable = config.getBool("writable", true);
+    bool visitable = config.getBool("visitable", true);
+    if (!config.getBool("list", visitable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::List);
+    }
+    if (!config.getBool("retrieve", visitable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::Retrieve);
+    }
+    if (!config.getBool("archive", writable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::Archive);
+    }
+    if (!config.getBool("wipe", writable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::Wipe);
+    }
+}
+
 ControlIdentifiers& ControlIdentifiers::operator|=(const ControlIdentifier& val) {
     value_ |= static_cast<value_type>(val);
     return *this;
--- a/src/fdb5/api/helpers/ControlIterator.cc
+++ b/src/fdb5/api/helpers/ControlIterator.cc
@@ -32,6 +32,33 @@ eckit::Stream& operator>>(eckit::Stream& s, ControlAction& a) {
 
 //----------------------------------------------------------------------------------------------------------------------
 
+std::ostream& operator<<(std::ostream& s, const ControlIdentifier& i) {
+    switch (i) {
+        case ControlIdentifier::None:
+            s << "None";
+            break;
+        case ControlIdentifier::List:
+            s << "List";
+            break;
+        case ControlIdentifier::Retrieve:
+            s << "Retrieve";
+            break;
+        case ControlIdentifier::Archive:
+            s << "Archive";
+            break;
+        case ControlIdentifier::Wipe:
+            s << "Wipe";
+            break;
+        case ControlIdentifier::UniqueRoot:
+            s << "UniqueRoot";
+            break;
+    }
+    s << "(" << static_cast<typename std::underlying_type<ControlIdentifier>::type>(i) << ")";
+    return s;
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+
 ControlIdentifierIterator::ControlIdentifierIterator(const ControlIdentifiers& identifiers) :
     value_(0), remaining_(identifiers.value_) {
 
--- a/src/fdb5/api/helpers/ControlIterator.cc
+++ b/src/fdb5/api/helpers/ControlIterator.cc
@@ -52,6 +52,9 @@ std::ostream& operator<<(std::ostream& s, const ControlIdentifier& i) {
         case ControlIdentifier::UniqueRoot:
             s << "UniqueRoot";
             break;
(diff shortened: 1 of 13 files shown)


src/fdb5/api/helpers/ControlIterator.cc
```diff
@@ -86,6 +113,24 @@ ControlIdentifiers::ControlIdentifiers(eckit::Stream& s) {
...
+    if (!config.getBool("list", visitable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::List);
+    }
+    if (!config.getBool("retrieve", visitable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::Retrieve);
+    }
+    if (!config.getBool("archive", writable)) {
+        value_ |= static_cast<value_type>(ControlIdentifier::Archive);
+    }
+    if (!config.getBool("wipe", writable)) {
```
caraghbiner, line 136: So, to be clear, we set this bit to 1 if:

- Wipe is set to false in the config.
- Wipe is not specified, but writable is explicitly false in config.

So, I assume value_ is going to store the set of actions we cannot do.

And we set to 0 if:

- Wipe is set to true.
- Wipe unspecified but writable is true
- If both wipe and writable not specified in config (i.e. by default)

And 0 means it is allowed i.e. wiping is by default enabled. Is this correct?

Worth adding a comment to this ctor that variables marked true in the config correspond to 0 in the ControlIdentifers, which is an aspect of the ControlIdentifiers I dislike.

src/fdb5/remote/server/CatalogueHandler.cc
```diff
@@ -44,7 +45,26 @@ namespace fdb5::remote {
 // ***************************************************************************************
 
 CatalogueHandler::CatalogueHandler(eckit::net::TCPSocket& socket, const Config& config) :
-    ServerConnection(socket, config), fdbControlConnection_(false), fdbDataConnection_(false) {}
+    ServerConnection(socket, config), fdbControlConnection_(false), fdbDataConnection_(false) {
+
+    eckit::net::IPAddress clientIPaddress{controlSocket_.remoteAddr()};
+
+    clientNetwork_ = "";
+    if (config_.has("networks")) {
```
caraghbiner, line 53: If I have a typo in my serverside config, e.g. I write `network` instead of `networks`, none of this logic is called, and (I think) everyone can wipe my data.

I'm inclined to say that for remote wipe it should default to being disabled.

caraghbiner (commented):
Pending test.

I am still worried about negligent / AI wipe commands from within a "trusted" network.

src/fdb5/api/helpers/ControlIterator.cc
tbkr: After looking at the enabled function in the ControlIdentifiers class, I think Chris is right and we should at least leave a comment describing that behavior.

If this isn't true, we should definitely leave a comment, describing the behavior.

I would default to non-wipeable, too.

tbkr (changes requested):
Would be good to documentation the behavior of the ControlIdentifiers and the mapping to configuration.

src/fdb5/api/helpers/ControlIterator.cc
```diff
@@ -86,6 +113,24 @@ ControlIdentifiers::ControlIdentifiers(eckit::Stream& s) {
     s >> value_;
 }
 
+ControlIdentifiers::ControlIdentifiers(const eckit::LocalConfiguration& config) : value_(0) {
```
simondsmart: I don't like this at all.

This ControlIdentifiers class is used in a number of places. It is used in the lock/unlock control functionality. And it is used here in the enable/disable logic. But those should both be logic which *uses* ControlIdentifiers rather than logic that is *in* ControlIdentifiers.

So, I'm really not at all convinced that the logic to extract enable/disable from config belongs inside this class. This is visible in the PR in that only *some* of the logic is here - the rest is in the CatalogueHandler.cc file. Yuck.

I also note Chris' comment below, such that the boolean logic is the inverse of what one would naively expect. We should probably fix that.

src/fdb5/remote/server/CatalogueHandler.cc
```diff
@@ -176,6 +209,11 @@ Handled CatalogueHandler::handleControl(Message message, uint32_t clientID, uint
 }
 
 
+void CatalogueHandler::checkIsEnabled(ControlIdentifier identifier) {
+    if (!controlIdentifiers_.enabled(identifier)) {
+        throw UnhandledOperationException(identifier);
```
simondsmart: This is not an *unhandled* operation (which draws to mind not-implemented, or similar). It is a *disabled* operation. A relative of PermissionDenied (the exception might derive from that)

src/fdb5/remote/server/CatalogueHandler.cc
simondsmart, line 148: Stats/Exist/Axes/List should also be considered for permissions.

src/fdb5/remote/server/CatalogueHandler.cc
```diff
@@ -343,6 +381,7 @@ void CatalogueHandler::axes(uint32_t clientID, uint32_t requestID, eckit::Buffer
 }
 
 void CatalogueHandler::wipe(uint32_t clientID, uint32_t requestID, eckit::Buffer&& payload) {
+    checkIsEnabled(ControlIdentifier::Wipe);
```
simondsmart: Is checkIsEnabled not called in handleControl()?

src/fdb5/remote/server/CatalogueHandler.cc
```diff
@@ -44,7 +45,26 @@ namespace fdb5::remote {
...
+    eckit::net::IPAddress clientIPaddress{controlSocket_.remoteAddr()};
+
+    clientNetwork_ = "";
+    if (config_.has("networks")) {
+        for (const auto& net : config_.getSubConfigurations("networks")) {
+            if (net.has("name") && net.has("netmask")) {
+                eckit::net::NetMask netmask{net.getString("netmask")};
+                if (netmask.contains(clientIPaddress)) {
+                    clientNetwork_ = net.getString("name");
+                    controlIdentifiers_ = ControlIdentifiers(net);
```
simondsmart: These apply in addition (?) to the logic in the contained FDB object?

simondsmart (changes requested):
In addition to the above, I want the ability to (fully) disable the wipe function on certain interfaces of the Store.

PR #268, opened 2026-04-16
FDB-664 added type for param strict matching + updated schema

danovaro:
### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-268


src/fdb5/types/TypeParam.cc
```diff
@@ -95,6 +100,16 @@ void TypeParam::print(std::ostream& out) const {
...
 
+class TypeParamStrict : public TypeParam {
+public:
+
+    TypeParamStrict(const std::string& name, const std::string& type) : TypeParam(name, type) {
+        strictMatching_ = true;
+    }
+};
+
+static TypeBuilder<TypeParamStrict> typeStrict("ParamID");
```
tbkr, line 111: If I understand the change correctly, this is registering a new Type "ParamID" which, if set in the schema, cause the matching to be strict. Maybe, as this is an added type to the paramId entry in the schema, rename it to StrictParamID or similar, to make clear that this is a strict matching, if stated in the schema. Currently it's somewhat opaque for the user what this does.

What is our overarching strategy here? Have a strict type for all keys which are able to be fuzzy matched? If so it may be worth it to implement the Strict type as a Decorator around `Type`, as this is not really a new type, but only an added characteristic to an already existing type.

tbkr (changes requested):
Left a comment about restructuring and potential use for other keys.

Set this to `request changes` but I may lack the bigger picture here, couldn't find infos in the corresponding ticket.

src/fdb5/types/TypeParam.cc
danovaro: I named it "ParamID" for consistency with Mars.
I acknowledge that the name is not self-explanatory, and I'm happy to change to "ParamStrict" or similar, if it improves clarity (at the cost of reduced consistency with Mars)
@simondsmart could you kindly comment on this point?

simondsmart: ParamID should be sufficient.

simondsmart (approved):

PR #269, opened 2026-04-16
FDB-638 remove Dist FDB

danovaro:
### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-269


src/fdb5/api/FDB.h
```diff
@@ -269,10 +269,6 @@ class FDB {
 
     // -------------- API management -----------------------------------------------------------------------------------
 
-    /// Returns the coniguration serialised into a string.
-    /// @return config as string.
-    const std::string id() const;
```
caraghbiner, line 274: I couldn't find it, but with double checking with Tobias that this isn't exposed in the python bindings.

caraghbiner (approved):

src/fdb5/api/FDB.h
tbkr: Just checked that this is not used in pyfdb.

tbkr (commented):
I checked the CICD and the pyfdb error are from the famous FDB-610 test. So this should be good in that regard.

I grep'd for dist fdb but couldn't find any more occurrences ;)

src/fdb5/api/FDB.h
simondsmart: I have made sure that it will not be present in the rust bindings as well. It should never have been user exposed.

simondsmart (approved):

PR #270, opened 2026-04-23
Fix incorrect data retrieval is special cases

Ozaq:
### Description

Now correctly retrieves data if non-chunked extension axis is defined.

Fixes: FDB 620

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-270

### DIFF
--- a/src/chunked_data_view/ChunkedDataViewImpl.cc
+++ b/src/chunked_data_view/ChunkedDataViewImpl.cc
@@ -20,6 +20,24 @@
 
 namespace chunked_data_view {
 
+namespace {
+
+size_t countFieldsForPart(const ViewPart& part, const std::vector<size_t>& chunkShape,
+                          const size_t extensionAxisIndex) {
+    size_t result = 1;
+    for (size_t i = 0; i < chunkShape.size() - 1; ++i) {
+        if (i == extensionAxisIndex) {
+            result *= part.shape()[i];
+        }
+        else {
+            result *= chunkShape[i];
+        }
+    }
+    return result;
+}
+
+}  // namespace
+
 ChunkedDataViewImpl::ChunkedDataViewImpl(std::vector<ViewPart> parts, size_t extensionAxisIndex) :
     parts_(std::move(parts)), extensionAxisIndex_(extensionAxisIndex) {
     shape_ = parts_[0].shape();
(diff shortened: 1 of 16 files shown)


src/chunked_data_view/ChunkedDataViewImpl.cc
tbkr, line 124: If no part matches as per off-by-one error, this silently fails. we should throw an error if a user tries to access anything beyond the extensions of the views.

src/chunked_data_view/ChunkedDataViewImpl.h
```diff
@@ -50,6 +50,19 @@ class ChunkedDataViewImpl : public ChunkedDataView {
         return result;
     }
 
+    size_t countFieldsForPart(const ViewPart& part) const {
```
tbkr: This is public and a helper only used in the NoChunking logic, if I see this correctly. Would even consider putting that in the cc as this is only a helper method.

src/chunked_data_view/GribExtractor.cc
```diff
@@ -55,15 +53,17 @@ size_t computeBufferIndex(const std::vector<Axis>& axes, const fdb5::Key& key) {
...
+    return chunked_data_view::index_mapping::axis_index_to_buffer_index(result, axes, extensionAxisIdx,
+                                                                        combinedExtSize, extensionOffset);
 }
 
-void GribExtractor::writeInto(const metkit::mars::MarsRequest& request,
-                              std::unique_ptr<ListIteratorInterface> list_iterator, const std::vector<Axis>& axes,
-                              const DataLayout& layout, float* ptr, size_t len, size_t expected_msg_count) const {
+size_t GribExtractor::writeInto(const metkit::mars::MarsRequest& request,
+                                std::unique_ptr<ListIteratorInterface> list_iterator, const std::vector<Axis>& axes,
+                                const DataLayout& layout, float* ptr, size_t len, size_t extensionAxisIdx,
```
tbkr, line 62: This function signature has grown quite a bit. I'd be a fan of introducing a stride object (or something with a better name) for reducing the amount of parameters needed here. Also the extensionAxisIdx is unsigned; using MAX_SIZE as the no-extension axis default is subpar compared to that parameter being a signed integer with -1 as default value, imho.

src/chunked_data_view/ChunkedDataViewImpl.cc
```diff
@@ -75,10 +75,28 @@ void ChunkedDataViewImpl::at(const std::vector<size_t>& chunkIndex, float* ptr,
         }
     }
 
+    if (!parts_[0].isAxisChunked(extensionAxisIndex_)) {
```
tbkr, line 96: Only the first part is considered to decide which logic to take for all parts. There is no ASSERT or validation that all parts agree on the chunking mode for the extension axis. 

We should probably check for inconsistencies between the individual parts and warn the user about the misconfiguration.

If a misconfigured view had mixed chunking across  parts, the wrong path would be silently chosen. 

I known that the builder prevents this for now, but the invariant is invisible at the point of use.

tbkr (changes requested):
So, big thanks for the PR. The overall implementation looks good to me besides the remarks I left.

I was more generally wondering: If we have a chunked axis, which is an extension axis, how do we handle that case? IIRC that would be the case for chunked date_time and multiple requests linked together (ofc you could formulate this as a single request).

Just to have something to talk about on monday.

And: It would be super lovely if we could have a picture/diagram showing the entire logic and nomenclature we introduced up to here. There are quite a lot of different terms, e.g. part, view, extensionAxis, the slicing, the relative offset within chunks etc.

Would be nice to have a schematic where the individual properties, e.g. for two combined parts are labelled just so that we clearly depict, what we are referring to.

That was lenghtly, nonetheless, thanks for looking into that. I really appreciate it. Took me also a bit to get my head around the individual details liked to the mapping <3

src/chunked_data_view/ChunkedDataViewImpl.h
Ozaq: Done

src/chunked_data_view/GribExtractor.cc
Ozaq: I am not going to change the parameters, id recommend a bit broader refactoring.

src/chunked_data_view/ChunkedDataViewImpl.cc
Ozaq: I think the invariants are properly established, i.e.
Shape = sum(partShape) and idx@ext_axis in interval [0, len(shape@ext_axis))

I can add a final throw at the end to signal breakage of invariants.

src/chunked_data_view/ChunkedDataViewImpl.cc
Ozaq: You are correct in your observation but I disagree in how to address it. IMO this indicates that this information should be pulled up into CDV. But I believe this is too much change in this PR.

tbkr (approved):
After talking to Kai he addressed the main points. The broader refactoring needs to be done in a separate PR.

PR #281, opened 2026-05-27
Feature/fdb remote deregister

danovaro:
### Description

fixes for local catalogue + remote stores:
- lazy initialisation of ReadLimiter
- fixed ClientConnectionRouter lifecycle 

improved client de-registration from the connection

small cleanup

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-281


src/fdb5/remote/client/Client.cc
```diff
@@ -69,8 +69,14 @@ void Client::refreshConnection() {
...
 
+void Client::deregister() {
+    if (!deregistered_.exchange(true)) {
+        connection_->remove(id_);
+    }
+}
+
 Client::~Client() {
-    connection_->remove(id_);
+    deregister();
```
simondsmart, line 79: Are there any exceptions that might be thrown during this deregister()? Is that safe, given we are in a destructor?

src/fdb5/remote/client/ReadLimiter.cc
```diff
@@ -18,21 +19,25 @@ namespace fdb5::remote {
 //----------------------------------------------------------------------------------------------------------------------
 namespace {
 ReadLimiter* instance_ = nullptr;
+std::once_flag initFlag_;
 }  // namespace
 
 bool ReadLimiter::isInitialised() {
```
simondsmart: I'm not sure what the value of isInitialised is.

src/fdb5/remote/client/ReadLimiter.cc
```diff
@@ -18,21 +19,25 @@ namespace fdb5::remote {
...
     return instance_ != nullptr;
 }
 
+void ReadLimiter::init(size_t memoryLimit) {
+    std::call_once(initFlag_, [memoryLimit] { instance_ = new ReadLimiter(memoryLimit); });
```
simondsmart: Note that this silently becomes a NOP if already initialised.

Is there a reason we can't just make the limit on the Readlimiter updatable? Then we cane use a `static ReadLimiter limiter` inside the instance() function rather than messing around with memory allocation/call_once.

src/fdb5/remote/client/ClientConnectionRouter.cc
```diff
@@ -110,8 +115,8 @@ void ClientConnectionRouter::deregister(ClientConnection& connection) {
 }
 
 ClientConnectionRouter& ClientConnectionRouter::instance() {
-    static ClientConnectionRouter router;
-    return router;
+    std::call_once(initFlag_, [] { instance_ = new ClientConnectionRouter(); });
```
simondsmart: I am missing something. Why are we changing the way we construct the static instance?

Are we having issues with the static instance being cleaned up too soon? if so we need a BIG comment in the places where this is done - not least to note that we will be leaking the object.

src/fdb5/remote/client/ClientConnectionRouter.cc
```diff
@@ -24,7 +25,11 @@ ConnectionError::ConnectionError(const eckit::net::Endpoint& endpoint) {
     reason(s.str());
     eckit::Log::status() << what() << std::endl;
 }
+
+fdb5::remote::ClientConnectionRouter* instance_ = nullptr;
```
simondsmart: Can this not be unique_ptr?

src/fdb5/remote/client/RemoteStore.cc
```diff
@@ -424,31 +410,13 @@ bool RemoteStore::handle(Message message, uint32_t requestID, eckit::Buffer&& pa
...
-                ASSERT(id != retrieveMessageQueues_.end());
-                id->second->emplace(std::make_pair(message, std::move(payload)));
-            }
+            std::lock_guard<std::mutex> lock(retrieveMessageMutex_);
+            auto id = retrieveMessageQueues_.find(requestID);
+            ASSERT(id != retrieveMessageQueues_.end());
+            id->second->emplace(std::make_pair(message, std::move(payload)));
             return true;
         }
         case Message::Error: {
```
simondsmart, line 426: I'm surprised to see that the interrupt() behaviour on the downstream message queues has gone. Is that correct. It feels like this creates a place where we could get stuck, with a thread waiting on the queues forever.

src/fdb5/toc/TocCatalogueWriter.h
```diff
@@ -110,6 +111,8 @@ class TocCatalogueWriter : public TocCatalogue, public CatalogueWriter {
 
     eckit::AutoUmask umask_;
     size_t archivedLocations_;
+
+    std::recursive_mutex archiveMutex_;
```
simondsmart: This is a bit of a smell. Why do we need a _recursive_ mutex here?

src/fdb5/toc/TocCatalogueWriter.cc
```diff
@@ -52,6 +52,7 @@ TocCatalogueWriter::~TocCatalogueWriter() {
 
 // selectIndex is called during schema traversal and in case of out-of-order fieldLocation archival
 bool TocCatalogueWriter::selectIndex(const Key& idxKey) {
+    std::lock_guard<std::recursive_mutex> lock(archiveMutex_);
```
simondsmart: Is this locking behaviour specific to the toc implementation, or is it something which belongs in the TocCatalogue base class?

src/fdb5/remote/Connection.cc
```diff
@@ -130,7 +148,7 @@ void Connection::write(const Message msg, const bool control, const uint32_t cli
 
 void Connection::error(std::string_view msg, uint32_t clientID, uint32_t requestID) const {
     eckit::Log::error() << "[clientID=" << clientID << ",requestID=" << requestID << "]  " << msg << std::endl;
-    write(Message::Error, false, clientID, requestID, msg.data(), msg.length());
+    write(Message::Error, true, clientID, requestID, msg.data(), msg.length());
```
simondsmart, line 151: This flip makes me nervous, and such that I would want to have a proper design discussion about it. Why are we changing whether (all) error messages are being sent on the control vs. data connection. That feels like a big change?

src/fdb5/remote/server/CatalogueHandler.cc
```diff
@@ -265,8 +265,6 @@ struct WipeHelper : public BaseHelper<CatalogueWipeState> {
     }
 
     WipeStateIterator apiCall(FDB& fdb, const FDBToolRequest& request) const {
-        // XXX: I'm inclined to say that in a multi-server scenario, unsafe wipe all is a bad idea.
```
simondsmart, line 268: Are we able to disable unsafe wipe all on the DE data bridge (particularly on LUMI) where we reindexed a large chunk of data?

simondsmart (changes requested):

src/fdb5/remote/Connection.cc
danovaro: Initially the control connection was blocking (no listener thread) and thus I had to use the data connection for the errors.
With the current implementation, we have the freedom to chose and the data connection could be busier with the large data transfer, so I considered this a potential improvement to the error latency.

src/fdb5/toc/TocCatalogueWriter.cc
danovaro: the race condition is generated by concurrent archival requests (generated by asynchronous replies sent by the stores involved in the archival).
in the read path, the visitor owns the  catalogue object, so we do not need any locking

src/fdb5/toc/TocCatalogueWriter.h
danovaro: we need to call selectIndex/currentIndex inside archive & index methods.
Following your suggestion, I've created private methods (not mutex protected) to be used inside archive & index

src/fdb5/remote/client/RemoteStore.cc
danovaro: fixed, thanks

src/fdb5/remote/client/ReadLimiter.cc
danovaro: removed

src/fdb5/remote/server/CatalogueHandler.cc
danovaro: reverted. Moving the changes to wipe permissions elsewhere

PR #285, opened 2026-06-04
Volfdb

danovaro:
### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-285

### DIFF
--- a/src/fdb5/api/FDB.cc
+++ b/src/fdb5/api/FDB.cc
@@ -21,6 +21,8 @@
 #include <string>
 #include <vector>
 
+#include <uuid/uuid.h>
+
 #include "eckit/config/Resource.h"
 #include "eckit/exception/Exceptions.h"
 #include "eckit/io/DataHandle.h"
@@ -39,7 +41,6 @@
 #include "fdb5/database/FieldLocation.h"
 #include "fdb5/database/Key.h"
 #include "fdb5/database/WipeCoordinator.h"
-#include "fdb5/database/WipeState.h"
 #include "fdb5/io/FieldHandle.h"
 #include "fdb5/io/HandleGatherer.h"
 #include "fdb5/message/MessageDecoder.h"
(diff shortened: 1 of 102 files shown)


src/fdb5/api/FDB.cc
```diff
@@ -66,24 +67,24 @@ FDB::FDB(FDB&&) = default;
...
 FDB& FDB::operator=(FDB&&) = default;
 
-void FDB::archive(eckit::message::Message msg) {
+void FDB::archive(eckit::message::Message msg, const std::string& tracingID) {
     fdb5::Key key = MessageDecoder::messageToKey(msg);
-    archive(key, msg.data(), msg.length());
+    archive(key, msg.data(), msg.length(), tracingID);
 }
-void FDB::archive(eckit::DataHandle& handle) {
+void FDB::archive(eckit::DataHandle& handle, const std::string& tracingID) {
```
caraghbiner, line 74: A passing comment: I would use `class TracingContext` or something similar rather than just a std::string. It seems plausible that this should be something that we might want to be extendable / structured.

src/fdb5/remote/client/ClientConnectionRouter.cc
```diff
@@ -112,14 +112,20 @@ void ClientConnectionRouter::deregister(ClientConnection& connection) {
...
-        instance_.reset(new ClientConnectionRouter());
-    }
-    return *instance_;
+    static ClientConnectionRouter router;
+    return router;
+    // std::lock_guard<std::mutex> lock(initMutex);
+    // if (!instance_) {
+    //     instance_.reset(new ClientConnectionRouter());
+    // }
+    // return *instance_;
```
caraghbiner, line 128: delete.

src/fdb5/api/local/MoveVisitor.cc
```diff
@@ -36,8 +36,8 @@ namespace local {
 //----------------------------------------------------------------------------------------------------------------------
 
 MoveVisitor::MoveVisitor(eckit::Queue<MoveElement>& queue, const metkit::mars::MarsRequest& request,
-                         const eckit::URI& dest) :
-    QueryVisitor<MoveElement>(queue, request), dest_(dest) {}
+                         const std::string& tracingID, const eckit::URI& dest) :
```
caraghbiner, line 39: Regarding tracing, I really think it would be better to use a thread_local static variable or similar rather than updating many many functions and passing a const string around everywhere.

src/fdb5/api/FDB.cc
```diff
@@ -354,6 +358,14 @@ void FDB::registerFlushCallback(FlushCallback callback) {
     internal_->registerFlushCallback(callback);
 }
 
+std::string FDB::generateTracingID(const std::string& prefix) {
```
jharwood0, line 361: I think we should format this as a W3C traceparent (`00-<trace-id>-<parent-id>-<flags>`), i'd keep it as a string we pass around, but format it inline with the standard https://www.w3.org/TR/trace-context/.

- it lines up with our codex (https://github.com/ecmwf/codex/blob/main/Guidelines/Observability.md#4-logging-standard) - traceId/spanId in logs for correlation, traceparent gives us those.
- we could optionally let the public fdb interface accept a traceparent input. Then any wrapping services or apps that already has a trace can pass theirs and our work hangs off it instead of floating as its own root.
- traceparent gives us depth context: trace-id stays constant for the whole request, parent-id is the caller's span-id, so hops can point back at their parent and you get a tree rather than a collection of logs with same id.

At the moment generateTracingID("archive") bakes "archive" into the id, but I'd argue that should be exposed as an attribute on a span context, not part of the identity. I know we're just talkign about Id structure at the moment but with a proper traceparent it would allow us to eventually capture fdb interactions something like this:

```
trace_id = 4bf92f3577b34da6a3ce929d0e0e4736
|
+- span "fdb.archive"         span_id=00f067aa0ba902b7   parent= (root)
   +- span "store.archive"     span_id=b7ad6b7169203331   parent=00f067aa0ba902b7   (data blob)
   +- span "catalogue.archive" span_id=9d6f2c1b4e7a8053   parent=00f067aa0ba902b7   (index entry)
```

PR #291, opened 2026-06-25
[python] add post compile to correct rpaths

tmi:
This makes the binaries distributed in the fdb5lib wheel contain the right relative rpaths. Not as strong as findlibs, but good enough for out of the box use

Test wheel publish: 
https://github.com/ecmwf/python-develop-bundle/actions/runs/28182996378/job/83477449822

I tested installing the wheel stack, importing pyfdb, and running the executables in the bin folder -- completely shallowly, just that "it does not crash on invocation". I tested **Linux only**

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-291

### DIFF
--- a/python/fdb5lib/post-compile.sh
+++ b/python/fdb5lib/post-compile.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+set -euo pipefail
+
+BASE=/tmp/fdb/target/fdb/bin
+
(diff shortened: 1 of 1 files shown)


simondsmart:
As discussed.

PR #320, opened 2026-08-05
FDB5->FDB in environment variables

caraghbiner:
### Description

- Support FDB_ form of the remaining FDB5_ environment variables.*
- Add docs for variables specified in tickets FDB-493/FDB-494/FDB-495/FDB-496/FDB-497/FDB-498/FDB-499/FDB-500/FDB-501/FDB-502.
- Remove dead code related to the variable FDB5_REMOTE_PROTOCOL_VERSION, which as far as I can tell has not been valid in the past 5 years.

*This excludes the pyfdb/findlibs use of FDB5_ variables for HOME and DIR.

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-320

### DIFF
--- a/docs/fdb/content/environment-variables.rst
+++ b/docs/fdb/content/environment-variables.rst
@@ -190,7 +190,7 @@ Default: ``true``.
 ``FDB_DATA_WRITE_QUEUE_LENGTH``
 -------------------------------
 
-Maximum length of the queue of pending data-write requests used by the remote client connection.
+Maximum number of requests in the queue of pending data-write requests on a remote client connection.
 
 Default: ``320``.
 
@@ -314,7 +314,8 @@ Indexing
 ------------------
 
 Overrides the index type used for new indexes. When left empty, FDB selects ``BTreeIndex`` for keys
-shorter than 8 characters and ``BTreeIndex64`` for longer keys.
+shorter than 8 characters and ``BTreeIndex64`` for longer keys. Accepted values are 
+``BTreeIndex``, ``BTreeIndex64``, ``PointDBIndex``, ``BTreeIndex4MB``.
 
 Default: unset (automatic selection).
 
@@ -325,9 +326,10 @@ Auxiliary data
 ``FDB_AUX_EXTENSIONS``
 ----------------------
 
-Set of filename extensions for auxiliary files associated with each archived field. For every data
-file written, FDB also considers auxiliary files sharing the field's path with one of these
-extensions appended.
+Set of accepted filename extensions for auxiliary files. These are files not produced directly by FDB, 
+but FDB will still consider them as part of the database for the purposes of wiping.
+Multiple extensions can be specified as a comma or space separated list, e.g.:
+   ``export FDB_AUX_EXTENSIONS="gribjump,foo,bar"``
 
 Default: ``gribjump``.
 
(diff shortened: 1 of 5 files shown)


docs/fdb/content/environment-variables.rst
```diff
@@ -50,4 +50,303 @@ overridden by ``FDB_CONFIG``.
...
+
+
+``FDB_ROOT``
+------------
+
+Used by the ``expver`` file space handler to force new ``expver`` data to be written to a specific
+root. The value must be one of the roots that support archival, otherwise an error is raised. If the
+``expver`` is already mapped to a different root, the existing mapping is kept and a warning is
+emitted.
+
```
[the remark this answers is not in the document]
caraghbiner: no

src/fdb5/toc/TocSerialisationVersion.cc
```diff
@@ -18,8 +18,11 @@ namespace fdb5 {
 static unsigned getUserEnvSerialisationVersion() {
 
     static unsigned fdbSerialisationVersion =
-        eckit::Resource<unsigned>("fdbSerialisationVersion;$FDB5_SERIALISATION_VERSION", 0);
-
+        eckit::Resource<unsigned>("fdbSerialisationVersion;$FDB_SERIALISATION_VERSION", 0);
+    if (!fdbSerialisationVersion) {
+        // backwards compatibility
+        fdbSerialisationVersion = eckit::Resource<unsigned>("$FDB5_SERIALISATION_VERSION", 0);
+    }
```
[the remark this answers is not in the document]
caraghbiner: too niche to care. explicitly setting FDB_SERIALISATION_VERSION=0 is unheard of.

docs/fdb/content/environment-variables.rst
```diff
@@ -50,4 +50,303 @@ overridden by ``FDB_CONFIG``.
...
 ``FDB5_CONFIG_FILE``
 --------------------
-Deprecated. Equivalent to ``FDB_CONFIG_FILE``. If both are specified, ``FDB_CONFIG_FILE`` takes precedence over ``FDB5_CONFIG_FILE``.
\ No newline at end of file
+Deprecated. Equivalent to ``FDB_CONFIG_FILE``. If both are specified, ``FDB_CONFIG_FILE`` takes precedence over ``FDB5_CONFIG_FILE``.
+
+
+.. note::
+
+   Several FDB environment variables historically used an ``FDB5_`` prefix. These are being
```
tbkr, line 58: I assume this means with regard to the FDB. For an user is probably isn't too transparent, what the difference to FDB5_HOME and FDB_HOME is and how/why this diverges. Also the FDB_HOME env var doesn't make an appearance? I wonder whether we should include this here, although I see that this was dropped as not being part of the scope.

src/fdb5/LibFdb5.cc
```diff
@@ -90,33 +90,7 @@ const std::set<std::string>& LibFdb5::auxiliaryRegistry() {
...
-
-    bool valid = check(user, false);
-    if (not valid) {
-        std::ostringstream msg;
-        msg << "Unsupported FDB5 remote protocol version " << user << " - supported: " << supportedStr() << std::endl;
-        throw eckit::BadValue(msg.str(), Here());
-    }
-    used_ = user;
-}
+RemoteProtocolVersion::RemoteProtocolVersion() : used_(defaulted()) {}
```
tbkr, line 93: I'm not knowledgeable enough, when it comes to the RemoteProtocolVersion, but is only the default a sensible option here? The user ability to specify it was entirely dropped, which may be fine. Maybe check, whether the check versions below is needed at all (or other code could be removed, as well). From the code it seems like 3 is the only valid value.

tbkr (approved):
Do we have a way of seeing whether this breaks anything? (speaking of the remote protocol version, as the rest is backwards compatible).

If not, how hard is it to generate the corresponding tests? I'd like to see those in the actual e2e tests eventually. Could you add a ticket for tracking those.

Would mention the FDB(5)_HOME situation, as I can this being slightly confusing for the end user.

Otherwise the changes look very nice!

caraghbiner:
From reading the code, the "remote protocol version" stuff logically does nothing: the function that checks the resource returns "0" instead of the value, and as far as I can tell it has been doing this for 5+ years, so not only does it predate the current iteration of the remote FDB, it has never functioned as a check. So, I'm not concerned about the risk of it breaking anything.

docs/fdb/content/environment-variables.rst
caraghbiner: This is a findlibs / pyfdb exclusive behaviour. I don't think the C++ has never looked at the FDB5_HOME variable.

caraghbiner: Oh FDB_HOME isn't here at all? Oops, obviously it should be. I guess because it is inferred from eckit so a ticket was never made based on the ctrl + F of FDB_*

caraghbiner: eh, a separate ticket I should think. There are possibly a few LIBNAME_ variables that we should document, but I don't want to do it now / in this PR

src/fdb5/LibFdb5.cc
caraghbiner: The function getUserEnvRemoteProtocol() returns 0 instead of the resource, which I assume is a bug, but appears to be a very old bug. So I'm removing the functionality entirely since we do not seem to have any intention of supporting user-configured protocol versions in the near term (Even if we support multiple protocol versions, I suspect it will be implemented by the client, not by the user config).

PR #322, opened 2026-08-06
[PyFDB] fdb-717: Improvements local dev setup pyfdb

tbkr:
### Description
Needed for local dev setup. No need to adjust the PYTHONPATH anymore. Installing the staging directories in build is sufficient now (after setting findlibs up for the local dev ofc).
### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-322


docs/pyfdb/technical_insights.rst
```diff
@@ -0,0 +1,192 @@
...
+   ecbuild_bundle( PROJECT metkit   GIT "https://github.com/ecmwf/metkit"   BRANCH develop UPDATE )
+   ecbuild_bundle( PROJECT fdb      GIT "https://github.com/ecmwf/fdb"      BRANCH develop UPDATE )
+
+   ecbuild_bundle_finalize()
+
+.. tip::
+
+   To work against a local checkout of any dependency instead of having ecbuild
+   clone it, replace the ``GIT`` / ``BRANCH`` arguments with
+   ``SOURCE_DIR <path-to-checkout>``. This is the typical workflow when you are
```
jharwood0:
```suggestion
   ``SOURCE <path-to-checkout>``. This is the typical workflow when you are
```

docs/pyfdb/technical_insights.rst
```diff
@@ -0,0 +1,192 @@
...
+   cd stack          # the directory that holds CMakeLists.txt
+   uv venv --python 3.11 # or any supported version
+   source .venv/bin/activate
+
+Install the Python build toolchain and ``PyFDB``'s runtime dependencies:
+
+.. code-block:: sh
+
+   uv pip install build setuptools wheel pybind11 \
+                  findlibs PyYAML pytest eccodes GitPython
```
jharwood0, line 82: Is eccodes python lib necessary here? It throws me a bit as I see eccodes is part of the build requirements for the fdb stack - but I imagine depending on which eccodes version we build against, this could drift from what is used here

jharwood0: I think you've mentioned ECCODES_PYTHON_USE_FINDLIBS + FINDLIBS_DISABLE_PACKAGE in trouble shooting section - but I guess these should be part of the main dev flow so we avoid installing these libs via pip at all?

docs/pyfdb/technical_insights.rst
```diff
@@ -0,0 +1,192 @@
...
+   To work against a local checkout of any dependency instead of having ecbuild
+   clone it, replace the ``GIT`` / ``BRANCH`` arguments with
+   ``SOURCE_DIR <path-to-checkout>``. This is the typical workflow when you are
+   modifying ``fdb`` or one of its dependencies alongside ``PyFDB``.
+
+Python Environment
+~~~~~~~~~~~~~~~~~~
+
+Create and activate a virtual environment in the ``stack`` root **before**
+running CMake. CMake's Python finder and pybind11 will both pick up the active
```
jharwood0, line 68: I'm not sure if this is just my setup - but fdb cmakelists fail to find pybind11 when resolving https://github.com/ecmwf/fdb/blob/develop/CMakeLists.txt#L136 - i think it may be because find_package(Python is defined below it https://github.com/ecmwf/fdb/blob/develop/CMakeLists.txt#L150

pytest.toml
```diff
@@ -0,0 +1,5 @@
+# Pytest configuration: Use importlib import mode to avoid test-package name
+# clashes with the real modules (e.g. tests/pyfdb vs pyfdb) caused by pytest's
+# default sys.path prepending.
+[pytest]
```
jharwood0: I was reading `pytest.toml` is new under pytest 9 - which got me thinking, do pin dev dependencies for this?

jharwood0 (approved):
pyfdb_bindings package changes makes sense 😃 

few minor comments / questions but no major blockers

docs/pyfdb/technical_insights.rst
tbkr: You need it for executing the tests, as the fixtures are filled with GRIB example files, generated by using eccodes. We would need to install the eccodes python bindings locally, as well. Let me check the feasability.

pytest.toml
tbkr: Good catch. I pushed this yesterday after the pyproject.toml had issues with our wheel-building pipeline.

docs/pyfdb/technical_insights.rst
tbkr: Good catch, fixed the ordering in there!

PR #325, opened 2026-08-07
cmake: pin python to build-time interpreter for editable installs

tmi:
The generated setup.py files for pyfdb and z3fdb are used both to build binary wheels and to allow editable installs directly from the build staging directory. Wheels already carry a python-version-specific tag enforced by pip, but the editable install path had no such protection: nothing prevented 'pip install -e' from succeeding under a different Python version than the one that built the compiled extension modules.

Use the VERSION_SUFFIX env var (already used to distinguish CD/publish builds from local dev builds) to pick the right python_requires:
  - VERSION_SUFFIX set (publish build, wheel path): keep the existing broad '>3.10' requirement.
  - VERSION_SUFFIX unset (local dev build, editable install path): pin python_requires to the exact major.minor of the Python cmake found via find_package(Python ...), using @Python_VERSION_MAJOR@ and @Python_VERSION_MINOR@ which are substituted at configure_file time.

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-325

### DIFF
--- a/cmake/pyfdb_setup.py.in
+++ b/cmake/pyfdb_setup.py.in
@@ -11,8 +11,15 @@ version_suffix = os.environ.get("VERSION_SUFFIX", "")
 if version_suffix:
     version_suffix = f".{version_suffix}"
     requires_extra = [f"fdb5lib==@fdb5_VERSION_STR@{version_suffix}"]
+    # Published wheels are binary and platform/version specific, but pip already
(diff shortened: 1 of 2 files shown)


tbkr (approved):
Rebased this to develop. This looks good to me.

PR #331, opened 2026-08-13
Fix/fdb uri

danovaro:
### Description

fixes FdbURIManager toString method (not including the endpoint)

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-331


src/fdb5/database/WipeState.cc
```diff
@@ -242,7 +246,10 @@ Store& StoreWipeState::store(const Config& config) const {
 
 void StoreWipeState::encode(eckit::Stream& s) const {
 
-    if (!signature_.isSigned()) {
+    static bool acceptUnsigned =
+        eckit::Resource<bool>("$FDB_ACCEPT_UNSIGNED_WIPE_STATE;fdbAcceptUnsignedWipeState", false);
```
caraghbiner, line 250: Does this bypass actually work? IIRC, the catalogue wipe has an assert that WIPE_SECRET is not empty.

Since the bypass requires setting an environment variable on both servers, I think one may as well just set the WIPE_SECRET instead.

src/fdb5/remote/server/StoreHandler.cc
```diff
@@ -488,7 +488,7 @@ void StoreHandler::finaliseWipeState(const uint32_t clientID, const uint32_t req
 
     // keep state for doWipeURIs
     if (doit) {
-        ASSERT(!unsafeAll);  // Until Im explicitly told otherwise, we dont support unsafeAll on remote fdb.
+        // ASSERT(!unsafeAll);  // Until Im explicitly told otherwise, we dont support unsafeAll on remote fdb.
```
caraghbiner, line 491: Not really related to the rest of this PR, so just double check that this is intentionally being committed, and then proceed you like.

src/fdb5/database/WipeState.cc
danovaro: in the local catalogue + remote store scenario, we might need to have the secret user readable, which makes it not so secret ;-)

this is the second half of a change in fix/FDB-714-select-partial-requests. Let me move it where it belongs

PR #346, opened 2026-09-02
FDB-723: remotefdb single-connection flush-blob ordering deadlock

mcakircali:
### Description

Tries to fix an deadlock in remote FDB when the client and server negotiate a single (shared) connection. Two CI tests are randomly timing out — fdb_test_remote_single_conn_stress (300s) and FDB-610 (600s).

On a single connection (shared a socket), client sends catalog (field-location) Blobs asynchronously via ClientConnection::dataWriteQueue_ (written later by the data-writer thread), but sends the blocking Flush msg directly.

The direct Flush may take over the queued Blobs. The server then reads Flush first, blocks in CatalogueHandler::flush() waiting, and cannot read the Blobs, which is a deadlock.

### Fix

Add barrier registry + failDataBarriers() for data writes so the catalog flush cannot arrive before Blobs.

`ClientConnection::flushDataWrites()` adds a barrier item (`std::promise<void>`) to the data writes queue and blocks until the writer thread reaches it and guarantees queued Blobs arrive before blocking flush.

### Related fixes

Add missing dataWriteQueue_ lock
Remove `throw` in thread; double throw bug

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-346


danovaro (approved):
thanks!

PR #349, opened 2026-09-07
FDB-739 fix SelectFDB config (userConfig, callback)

danovaro:
### Description

fixed SelectFDB config & callback propagation 

- it is passing userConfig to the SelectFDB lanes
- it is creating a shared object holding the callback, thus registering the callbacks with the parent object makes them available for all the lanes
- test config is now using selectFDB

### Contributor Declaration

🌈🌦️📖🚧 Documentation FDB 🚧📖🌦️🌈
https://sites.ecmwf.int/docs/fdb/pull-requests/PR-349

### DIFF
--- a/src/fdb5/api/FDBFactory.cc
+++ b/src/fdb5/api/FDBFactory.cc
@@ -62,11 +62,11 @@ bool FDBBase::enabled(const ControlIdentifier& controlIdentifier) const {
 }
 
 void FDBBase::registerFlushCallback(FlushCallback callback) {
-    callbacks_->flushCallback_ = callback;
+    callbacks_->flushCallback_ = std::move(callback);
 }
 
 void FDBBase::registerArchiveCallback(ArchiveCallback callback) {
-    callbacks_->archiveCallback_ = callback;
+    callbacks_->archiveCallback_ = std::move(callback);
 }
 
 void FDBBase::setCallbacks(std::shared_ptr<Callbacks> callbacks) {
--- a/src/fdb5/api/FDBFactory.cc
+++ b/src/fdb5/api/FDBFactory.cc
@@ -28,7 +28,8 @@ namespace fdb5 {
 //----------------------------------------------------------------------------------------------------------------------
 
 
-FDBBase::FDBBase(const Config& config, const std::string& name) : name_(name), config_(config) {
+FDBBase::FDBBase(const Config& config, const std::string& name) :
+    name_(name), config_(config), callbacks_(std::make_shared<Callbacks>()) {
 
     bool writable = config.getBool("writable", true);
     bool visitable = config.getBool("visitable", true);
@@ -58,6 +59,19 @@ const Config& FDBBase::config() const {
 
 bool FDBBase::enabled(const ControlIdentifier& controlIdentifier) const {
     return controlIdentifiers_.enabled(controlIdentifier);
+}
+
+void FDBBase::registerFlushCallback(FlushCallback callback) {
+    callbacks_->flushCallback_ = callback;
+}
+
+void FDBBase::registerArchiveCallback(ArchiveCallback callback) {
+    callbacks_->archiveCallback_ = callback;
+}
+
+void FDBBase::setCallbacks(std::shared_ptr<Callbacks> callbacks) {
+    ASSERT(callbacks);
+    callbacks_ = std::move(callbacks);
 }
 
 FDBFactory& FDBFactory::instance() {
--- a/src/fdb5/api/LocalFDB.cc
+++ b/src/fdb5/api/LocalFDB.cc
@@ -67,7 +67,7 @@ void LocalFDB::archive(const Key& key, const void* data, size_t length) {
         std::lock_guard lock(mutex_);
         if (!archiver_) {
             LOG_DEBUG_LIB(LibFdb5) << *this << ": Constructing new archiver" << std::endl;
-            archiver_ = std::make_unique<Archiver>(config_, archiveCallback_);
+            archiver_ = std::make_unique<Archiver>(config_, callbacks_->archiveCallback_);
         }
         archiver = archiver_.get();
     }
@@ -172,11 +172,11 @@ void LocalFDB::flush() {
     }
     if (archiver) {
         archiver->flush();
-        flushCallback_();
+        callbacks_->flushCallback_();
     }
     else if (reindexer) {
         reindexer->flush();
-        flushCallback_();
+        callbacks_->flushCallback_();
     }
 }
 
--- a/src/fdb5/api/SelectFDB.cc
+++ b/src/fdb5/api/SelectFDB.cc
@@ -43,7 +43,9 @@ static FDBBuilder<SelectFDB> selectFdbBuilder("select");
 
 //----------------------------------------------------------------------------------------------------------------------
 
-SelectFDB::FDBLane::FDBLane(const eckit::LocalConfiguration& config) : config_(config), fdb_(nullptr) {
+SelectFDB::FDBLane::FDBLane(const eckit::LocalConfiguration& config, const eckit::Configuration& userConfig,
+                            std::shared_ptr<Callbacks> callbacks) :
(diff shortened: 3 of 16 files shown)


src/fdb5/api/LocalFDB.cc
```diff
@@ -67,7 +67,7 @@ void LocalFDB::archive(const Key& key, const void* data, size_t length) {
         std::lock_guard lock(mutex_);
         if (!archiver_) {
             LOG_DEBUG_LIB(LibFdb5) << *this << ": Constructing new archiver" << std::endl;
-            archiver_ = std::make_unique<Archiver>(config_, archiveCallback_);
+            archiver_ = std::make_unique<Archiver>(config_, callbacks_->archiveCallback_);
```
caraghbiner: `Archiver` stores its callback as `const ArchiveCallback& callback_`, i.e. a reference into the current `Callbacks` object. If `setCallbacks()` were ever called on this `LocalFDB` *after* the archiver has been created, the old `Callbacks` object could be destroyed, leaving the archiver with a dangling reference (and it would in any case keep invoking the old callback).

Today this can't happen: `setCallbacks()` is only invoked by `SelectFDB::FDBLane::get()` immediately after `build()`, before any `archive()`. But the invariant is implicit. Consider either passing the `std::shared_ptr<Callbacks>` down into `Archiver`, or overriding `setCallbacks()` in `LocalFDB` with `ASSERT(!archiver_)` to make the precondition explicit.

src/fdb5/api/SelectFDB.cc
```diff
@@ -56,6 +58,7 @@ bool SelectFDB::FDBLane::matches(const T& vals, Matcher::MatchMissingPolicy matc
 FDBBase& SelectFDB::FDBLane::get() {
     if (!fdb_) {
         fdb_ = FDBFactory::instance().build(config_);
+        fdb_->setCallbacks(callbacks_);
```
caraghbiner: This replaces the `Callbacks` object the sub-FDB created in its own `FDBBase` constructor, silently discarding anything that may have been registered on it in between. That's fine today because `FDBFactory::build()` does not fire `LibFdb5::constructorCallback()` (only `FDB::FDB` does), but it's an ordering assumption worth a short comment here — or, longer term, consider passing the shared `Callbacks` through the factory so a lane FDB never owns a temporary registry.

src/fdb5/api/helpers/Callback.h
```diff
@@ -34,17 +34,19 @@ static const ConstructorCallback CALLBACK_CONSTRUCTOR_NOOP = [](auto&&...) {};
 
 // -------------------------------------------------------------------------------------------------
 
+struct Callbacks {
```
caraghbiner: Two things worth documenting on `Callbacks`:

1. Registration now mutates state shared by all lanes with no synchronization, and `LocalFDB`'s `Archiver` can invoke `archiveCallback_` from its own flush/archive path — so callbacks must be registered before the FDB is used. A one-line comment stating that would help.
2. There is a single slot per callback type for the whole FDB tree: registering a second callback (on the parent or on any lane) silently replaces the first everywhere. Previously each sub-FDB had its own independent slot. I believe this is the intended semantics of the fix, but it's a behaviour change worth calling out in the docstring.

src/fdb5/rules/SelectMatcher.cc
```diff
@@ -48,13 +52,22 @@ class KeyAccessor : public metkit::mars::RequestLike {
...
+    }
     if (config.has("excludes")) {
         for (const auto& ex : config.getStringVector("excludes")) {
             excludes_.push_back(Matcher(ex, Matcher::Policy::All));
         }
+        found = true;
+    }
+    if (!found) {
+        std::ostringstream ss;
+        ss << "SelectMatcher: no select or excludes rule found in config: " << config;
```
caraghbiner: Nice improvement — previously a lane without `select` would hit metkit's opaque `BadValue("Empty match expression")`, and excludes-only lanes (as now used in `tests/fdb/etc/fdb/config.yaml`) were impossible. The explicit error with the offending config is much friendlier.

src/fdb5/api/FDBFactory.cc
```diff
@@ -60,6 +61,19 @@ bool FDBBase::enabled(const ControlIdentifier& controlIdentifier) const {
     return controlIdentifiers_.enabled(controlIdentifier);
 }
 
+void FDBBase::registerFlushCallback(FlushCallback callback) {
+    callbacks_->flushCallback_ = callback;
```
caraghbiner: Nit: `callback` is taken by value in both register methods — `callbacks_->flushCallback_ = std::move(callback);` would avoid an extra `std::function` copy. Same below for the archive callback.

tests/fdb/etc/fdb/config.yaml
```diff
@@ -1,15 +1,26 @@
...
-  roots:
-  - path: @PROJECT_BINARY_DIR@/tests/fdb/root
-    writable: true
-    visit: true
-  - path: @PROJECT_BINARY_DIR@/tests/fdb/root2
-    wipe: true
-    archive: false
-    list: true
-    retrieve: true
+type: select
```
caraghbiner: Two small points on the new test config:

1. The `class=rd` lane's root now has no attribute flags (relying on `writable`/`visit` defaults) and rd data no longer sees `root2`'s `wipe`/`list`/`retrieve` settings — worth double-checking no existing test archives `class=rd` and depends on the root2 behaviour.
2. `schema` is duplicated in both lanes; `SelectFDB` injects a top-level `schema` into lanes that don't define one, so it could be set once at the top.

src/fdb5/remote/server/StoreHandler.h
```diff
@@ -62,6 +62,9 @@ class StoreHandler : public ServerConnection, public CallbackRegistry {
     Store& store(uint32_t clientID, const Key& dbKey);
     Store& getStore(uint32_t clientID, const eckit::URI& uri);
 
+    void registerFlushCallback(FlushCallback callback) override;
```
caraghbiner: Minor: these overrides are in the private section, so callbacks can now only be registered via a `CallbackRegistry&` (which is how the constructor-callback plugins, e.g. gribjump, use it). If that restriction is intentional, fine — just flagging that direct calls on a concrete `StoreHandler` no longer compile.

src/fdb5/api/FDBFactory.cc
```diff
@@ -60,6 +61,19 @@ bool FDBBase::enabled(const ControlIdentifier& controlIdentifier) const {
...
+}
+
+void FDBBase::registerArchiveCallback(ArchiveCallback callback) {
+    callbacks_->archiveCallback_ = callback;
```
caraghbiner: I believe both of these could uses `std::move` to avoid an extra copy e.g. `callbacks_->flushCallback_ = std::move(callback);`

src/fdb5/rules/SelectMatcher.cc
```diff
@@ -48,13 +52,22 @@ class KeyAccessor : public metkit::mars::RequestLike {
 
 // ------------------------------------------------------------------------------------------------------
 
-SelectMatcher::SelectMatcher(const eckit::LocalConfiguration& config) :
-    select_{Matcher(config.getString("select", ""), Matcher::Policy::Any)} {
-
+SelectMatcher::SelectMatcher(const eckit::LocalConfiguration& config) {
+    bool found = false;
+    if (config.has("select")) {
```
caraghbiner, line 57: Woah, does this mean we are supporting lanes without any "select" entry? Lanes which are defined only be exclude? Is that related to the userConfig not being propagating / callbacks not being registered, or a bonus change?

caraghbiner (approved):
Thanks, one trivial comment and one question about what it means for a lane to not have a `select` entry.

src/fdb5/rules/SelectMatcher.cc
danovaro: It helps creating a set of not overlapping lanes that covers all the cases.
In this PR I've created a simple testing config.yaml with 2 lanes: `class=rd`, and then `class != rd`

